PR #1321, opened 2019-12-20
Allow composer install and including through rector.yaml in other project

MK-42:
To be able to use this package, I had to do some adjustments


sabbelasichon:
@MK-42 Thanks a lot.

PR #1363, opened 2020-02-10
[TASK] Add TCA migration rector

sabbelasichon:


tests/config/typo3_rectors.yaml
```diff
@@ -1,4 +1,5 @@
 imports:
+    - { resource: "../../config/tca_migration.yaml" }
```
Tuurlijk: Should we split these up by TYPO3 version?

sabbelasichon:
Is handled by #1522

PR #1394, opened 2020-02-23
[TASK] Fix CascadeAnnotationRector for @cascade remove annotations

MK-42:
Many projects use `@cascade remove` instead of `@cascade` alone.

This patch fixes rewriting these occurences of the annotation to the new namespaced versions.


MK-42:
thanks @sabbelasichon !

PR #1406, opened 2020-04-27
Improve coding standard + put Rector into work

TomasVotruba:
- added [PSR-12 (PSR-2 is deprecated since last year](https://www.php-fig.org/psr/psr-12/)
- added PHP 7.0 set + strict types
- added PHP 7.1 set

Let me know if this is ok for you.


TomasVotruba:
I also put Rector to work ;)

Let me know if some coding standard rule doesn't suite your needs. We can ignore it.

Ready to merge :+1:

TomasVotruba:
:partying_face:

PR #1463, opened 2020-07-23
[TASK] Apply renamed RenameMethodCallRector to config

utrotzek:
As mentioned in the ChangeLog of the rector project
the rector RenameMethodCallRector has been renamed to
RenameMethodRector since version 0.7.49

See: https://github.com/rectorphp/rector/blob/0bef4d82622b3ebd2ad7fc474ef117a4a56d2487/CHANGELOG.md#v0749---2020-07-21


utrotzek:
I'm not sure why the checks are failing, but it seems that the problem lays in the basic infrastructure and not in the change itself.

sabbelasichon:
@utrotzek  Thanks. Sorry for the late response. Solved by #1467

Issue #1452, opened 2020-05-06
Deprecation: #90800 - GeneralUtility::isRunningOnCgiServerApi



sabbelasichon:
Deprecation: #90800 - GeneralUtility::isRunningOnCgiServerApi

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Deprecation-90800-GeneralUtilityisRunningOnCgiServerApi.html

.. include:: ../../Includes.txt

=============================================================

Deprecation: #90800 - GeneralUtility::isRunningOnCgiServerApi

=============================================================

See :issue:`90800`

Description

===========

The lowlevel API method :php:`GeneralUtility::isRunningOnCgiServerApi()` which detects if

the current PHP is executed via a CGI wrapper script ("SAPI", see https://www.php.net/manual/en/function.php-sapi-name.php) has been

moved to the Environment API and is now available via :php:`Environment::isRunningOnCgiServer()`.

Impact

======

Calling the method from :php:`GeneralUtility` will trigger a PHP :php:`E_USER_DEPRECATED` error.

Affected Installations

======================

Any TYPO3 installation with an extension using this PHP method, which will happen only in rare circumstances.

Migration

=========

Use the new method :php:`Environment::isRunningOnCgiServer()` instead, which works exactly the same.

.. index:: PHP-API, FullyScanned, ext:core

PR #1478, opened 2020-09-14
[TASK] Use class Environment for method isRunningOnCgiServer

sabbelasichon:
Resolves: #1452


tmotyl:
Is documentation updated automatically, or should a note about this rector land there too?

sabbelasichon:
@tmotyl The documentation is generated automatically for custom Rectors but not for the configured ones.

PR #1481, opened 2020-09-25
Migrate famous TYPO3_DB to doctrine dbal

sabbelasichon:


tmotyl:
looks good, we can improve it further after merge

PR #1521, opened 2020-10-18
Deprecation: #89127 - Cleanup RecordHistory handling

### LINKED ISSUE #1291: Deprecation: #89127 - Cleanup RecordHistory handling
Deprecation: #89127 - Cleanup RecordHistory handling

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.1/Deprecation-89127-CleanupRecordHistoryHandling.html

====================================================

Deprecation: #89127 - Cleanup RecordHistory handling

====================================================

See :issue:`89127`

Description

===========

The following properties of the :php:`\TYPO3\CMS\Backend\History\RecordHistory` class have been marked as deprecated:

* :php:`changeLog`

* :php:`lastHistoryEntry`

The properties are now protected and have a public getter function.

The following public methods of the :php:`\TYPO3\CMS\Backend\History\RecordHistory` class have changed visibility from public to protected:

* :php:`getHistoryEntry()`

* :php:`getHistoryData()`

The following methods of the :php:`\TYPO3\CMS\Backend\History\RecordHistory` class have been marked as deprecated:

* :php:`createChangeLog()`, use :php:`getChangeLog()` instead

* :php:`shouldPerformRollback()`

* :php:`getElementData()`, use :php:`getElementInformation()` instead

* :php:`performRollback()`, use :php:`RecordHistoryRollback::performRollback()` instead

* :php:`createMultipleDiff()`, use :php:`getDiff()` instead

* :php:`setLastHistoryEntry()`, use :php:`setLastHistoryEntryNumber()` instead

Impact

======

Accessing these properties and methods directly will trigger a PHP :php:`E_USER_DEPRECATED` error.

Affected Installations

======================

TYPO3 installations with custom extensions or TypoScript directly

accessing these properties and methods.

Migration

=========

Use the mentioned alternative methods and new classes.

.. index:: Backend, PHP-API, FullyScanned, ext:backend

verdure-peuser:
Fixes #1291 

New rector SubstituteDeprecatedrecordHistoryMethodsRector

removes removed/deprecated properties/methods
migrates to alternative methods


sabbelasichon:
@verdure-peuser If you don´t mind i would just take the configurable rectors. I would not remove the nodes,  user intention is unclear and i want to be as defensive as possible. User has to take what to do next.

verdure-peuser:
Alright. So i remove all parts besides the renaming/configurable parts ? In this case, why not split it up ?
 
So this part works besides the configurable isn't working (since you explained to me the confogurable part isn't loaded in the testcase and you'd prefer to not use it that way). 
Meaning we would have the rector for removing (and testing only this), and the configurable part which has a test with renaming. This would contain all required changes described in the rst.
Otherwise if we would keep it in one single rector, we would have to deal with the renaming on our own (besides we have an option/configurable rector for it).
The only part which isn't shining is, the renaming only happens in the configuration, no specific rector for it.

What do you think ?

PR #1522, opened 2020-10-20
[TASK] Add TCA migration

sabbelasichon:
- Add TCA migration using the migrators shipped by the TYPO3 core.
- It is only possible for files in TCA returning the configuration


tmotyl:
This is aweeesome!

Issue #1456, opened 2020-05-06
Deprecation: #90803 - ObjectManager::get in Extbase context



sabbelasichon:
Deprecation: #90803 - ObjectManager::get in Extbase context

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.4/Deprecation-90803-DeprecationOfObjectManagergetInExtbaseContext.html

.. include:: ../../Includes.txt

===========================================================

Deprecation: #90803 - ObjectManager::get in Extbase context

===========================================================

See :issue:`90803`

Description

===========

To help understand the deprecation of :php:`$objectManager->get(Service::class)` let's first have a look at its domain: Dependency Injection

and its history as well as the culprits to deal with.

With the introduction of Extbase over one decade ago, a lot of modern software development paradigms have been introduced into TYPO3.

One of that paradigms is Dependency Injection (DI) which is an approach of handling dependencies different than the one the TYPO3 core followed ever since.

Given there is an EmailService class, which is responsible for sending emails, the usual approach of creating such a service was to create it

the moment it was needed. TYPO3 never used the :php:`new` keyword to create new objects, but :php:`GeneralUtility::makeInstance()`, which pretty much does the same thing.

So, one approach of creating dependencies is creating them in the current scope where the dependency is needed.

.. tip::

   As a rule of thumb, you can remember the following:

   Whenever you are creating dependencies yourself with :php:`new` or :php:`GeneralUtility::makeInstance()`, you are not using Dependency Injection.

Extbase introduced the concept of Dependency Injection (DI) which means, that all dependencies are declared in a way, that the dependency chain is known before runtime.

The most common way of implementing DI is to declare dependencies as constructor arguments. This means, in the scope of the current class, all dependencies are made visible as constructor arguments.

As those dependencies need to be created outside the current scope, a service container implementation is responsible for the creation and management of service instances.

Then, instead of calling :php:`new Service(...)`, the container needs to be queried for the needed service, e.g. by calling :php:`$container->get(Service::class)`.

This also assures that the container provide the requested services with their dependencies, as they are created the same way.

There is an service container in Extbase but it's not exposed to the public. Instead, there is the :php:`ObjectManager` class, which acts as a proxy for the container and also has a :php:`get` method, to query instances of services.

Exactly that :php:`get()` method is now deprecated in the extbase context because it should never be called directly.

[...]

PR #1532, opened 2020-10-29
[TASK] Move calls ObjectManager->get to constructor injection

sabbelasichon:
Resolves: #1456


sabbelasichon:
This is just a proof of concept. As far as i can see, this is not really reliable solvable and produces more harm than good. Task has to be done carefully and manually.

PR #1548, opened 2020-11-02
Create phar file for typo3-rector

sabbelasichon:


sabbelasichon:
@Tuurlijk I have done something very simple to create a phar file out of the project. This seems to work. What do you think?

PR #1602, opened 2020-11-15
Issue 423

sabbelasichon:


src/Rector/v8/v0/TimeTrackerInsteadOfNullTimeTrackerRector.php
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        if (! $this->isMethodStaticCallOrClassMethodObjectType($node, GeneralUtility::class)) {
+            return false;
+        }
+
+        return $this->isName($node->name, 'makeInstance');
+    }
+
+    private function isObjectMangerCall(Node $node): bool
```
ayacoo, line 142: Manager

captainhook.json
```diff
@@ -16,17 +16,17 @@
                 "conditions": []
             },
             {
-                "action": "vendor/bin/ecs check --ansi",
+                "action": "composer run check-style",
```
ayacoo, line 19: Do the composer changes need to be made in this commit? So you have a dependency on the rector? Otherwise it would be a separate commit for me.

sabbelasichon: It is a separate commit.

PR #1615, opened 2020-11-19
[TASK] GeneralUtility::getUrl RequestHeaders as associative array

ayacoo:


ayacoo:
Need some help ;-)

The first 3 lines from the test are green. The argument is directly overwritten, if present.

The variable $headers (line 4 and 5 from the test) is a bit difficult. How do you deal with such a case? You can check for an assign of course, but what if the headers are e.g. a class constant or a class variable? I'm not sure if Rector will reach his limits here again.

ayacoo:
see https://github.com/sabbelasichon/typo3-rector/pull/1618

Issue #1139, opened 2019-11-04
Deprecation: #85462 - Signal 'hasInstalledExtensions'



sabbelasichon:
Deprecation: #85462 - Signal 'hasInstalledExtensions'

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.4/Deprecation-85462-SignalHasInstalledExtensions.html

.. include:: ../../Includes.txt

=====================================================

Deprecation: #85462 - Signal 'hasInstalledExtensions'

=====================================================

See :issue:`85462`

Description

===========

The usage of signal :php:`hasInstalledExtensions` of class

:php:`\TYPO3\CMS\Extensionmanager\Service\ExtensionManagementService`

has been marked as deprecated and will be removed in TYPO3 v10.

The signal is a duplication of :php:`afterExtensionInstall` that is also emitted during

extension installation.

Impact

======

Slots of this signal will get executed in TYPO3 v9 but will be abandoned with TYPO3 v10.

Affected Installations

======================

Extensions that register slots for the signal :php:`hasInstalledExtensions` of class

:php:`\TYPO3\CMS\Extensionmanager\Service\ExtensionManagementService`.

Migration

=========

Extensions should use the signal :php:`afterExtensionInstall` of class

:php:`\TYPO3\CMS\Extensionmanager\Utility\InstallUtility` instead which is emitted after an

extension has been installed.

.. index:: Backend, LocalConfiguration, PHP-API, NotScanned, ext:extensionmanager

tomasnorre:
I'll be trying my luck with this one :)

tomasnorre:
Looking for some inspiratation and saw this: 
https://github.com/sabbelasichon/typo3-rector/blob/master/src/Rector/v9/v4/UseSignalTablesDefinitionIsBeingBuiltSqlExpectedSchemaServiceRector.php#L72

Isn't that a mistake that the two use statements are on same line?

sabbelasichon:
It is more a unpretty formatting issue. Nothing problematic codewise. If you like you could fix this with your PR.

tomasnorre:
Will see if I can have it done during the weekend.

PR #1624, opened 2020-11-21
[TASK] Use signal afterExtensionInstall instead of hasInstalledExtensions (#1139)

tomasnorre:
Resolves: #1139

### DIFF
--- a/phpstan-baseline.neon
+++ b/phpstan-baseline.neon
@@ -10,3 +10,7 @@ parameters:
 			count: 1
 			path: src/Set/Typo3RectorSetProvider.php
 
+		-
+			message: '#Property PhpParser\\Node\\Arg\:\:\$value \(PhpParser\\Node\\Expr\) does not accept PhpParser\\Node\\Arg#'
+			count: 1
+			path: src/Rector/v9/v4/UseSignalAfterExtensionInstallInsteadOfHasInstalledExtensionsRector.php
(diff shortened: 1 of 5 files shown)


phpstan-baseline.neon
```diff
@@ -10,3 +10,7 @@ parameters:
 			count: 1
 			path: src/Set/Typo3RectorSetProvider.php
 
+		-
```
tomasnorre, line 13: I had to add this to have it pass the PHPStan. I could not figure out how to assign the value to the ` $node->args[1]->value` without using the `$this->createArg()` 

https://github.com/sabbelasichon/typo3-rector/pull/1624/files#diff-42ad5bdb72bdf2ba6a8115a562bc10ea4a17046fdea25bb65a9c2973616bd997R48

tomasnorre: I found this as an option
```php
$node->args[1]->value = new Node\Arg(BuilderHelpers::normalizeValue('afterExtensionInstall'));
```
but the `BuilderHelps` is marked as `@internal`

sabbelasichon: @tomasnorre Either you assign the new Arg this way: $node->args[1] = $this->createArg('afterExtensionInstall'). Behind the scenes Rector is using the BuilderHelpers::normalizeValue method or you assign a new value that way: $node->args[1]->value = new Node\Scalar\String_('afterExtensionInstall'); You can also use the BuilderHelpers::normalizeValue('afterExtensionInstall') method. 

Afterwards you can remove the line in the phpstan-baseline.neon.

sabbelasichon:
@tomasnorre Really good for the first one. Keep it up.

phpstan-baseline.neon
tomasnorre: I liked this option the most, as you see that it's `$node->args[1]->value` being set, but PHPStan complains about partials use: 
`$node->args[1]->value = new Node\Scalar\String_('afterExtensionInstall');`

So I went for ` $node->args[1] = $this->createArg('afterExtensionInstall')` instead.

Issue #1253, opened 2019-11-04
Deprecation: #88366 - Default caching framework cache names changed



sabbelasichon:
Deprecation: #88366 - Default caching framework cache names changed

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.0/Deprecation-88366-DefaultCachingFrameworkCacheNamesChanged.html

.. include:: ../../Includes.txt

===================================================================

Deprecation: #88366 - Default caching framework cache names changed

===================================================================

See :issue:`88366`

Description

===========

TYPO3's internal Caching Framework has several caches already shipped with TYPO3 Core.

The caches have been renamed for convenience and for newcomers to overcome another "speciality"

of TYPO3 which was due to legacy and integration reasons back in TYPO3 4.3 when the Caching

Framework was introduced.

The following caches have been renamed:

* `cache_core` => `core`

* `cache_hash` => `hash`

* `cache_pages` => `pages`

* `cache_pagesection` => `pagesection`

* `cache_runtime` => `runtime`

* `cache_rootline` => `rootline`

* `cache_imagesizes` => `imagesizes`

The caches should now be accessed via :php:`$cacheManager->getCache('core')` instead of

:php:`$cacheManager->getCache('cache_core')` - without the ``cache_`` prefix.

In addition, when the DatabaseBackend cache is used, the database tables do not have the :sql:`cf_`

prefix anymore, making it clearer for integrators and developers what the caches mean.

Impact

======

When accessing the cache with a "cache" prefix, a PHP :php:`E_USER_DEPRECATED` error is triggered.

Affected Installations

======================

Any TYPO3 extension using the caching framework with the ``cache_`` prefix.

Migration

=========

Remove the ``cache_`` prefix from the callers code.

In addition, run through the Database Table Analyzer of the Configuration module to

re-create any database tables of the Caching Framework.

.. index:: Database, LocalConfiguration, NotScanned, ext:core

tomasnorre:
I think I can take care of this one too. Quite similar to what I just did. Just not static functions. Will wait for feedback first on the other, not to do something "wrong" twice.

Issue #1059, opened 2019-11-04
Deprecation: #85012 - GetValidationResults of Argument:class and Arguments::class



sabbelasichon:
Deprecation: #85012 - GetValidationResults of Argument:class and Arguments::class

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.3/Deprecation-85012-OnlyValidateMethodParamsIfNeeded.html

.. include:: ../../Includes.txt

=================================================================================

Deprecation: #85012 - GetValidationResults of Argument:class and Arguments::class

=================================================================================

See :issue:`85012`

Description

===========

The following public methods have been marked as deprecated:

* :php:`TYPO3\CMS\Extbase\Mvc\Controller\Argument::getValidationResults()`

* :php:`TYPO3\CMS\Extbase\Mvc\Controller\Arguments::getValidationResults()`

Impact

======

Calling the method will trigger a PHP :php:`E_USER_DEPRECATED` error.

Affected Installations

======================

Extensions that call any of the methods. The extension scanner will find usages.

Migration

=========

Use the following methods instead:

* :php:`TYPO3\CMS\Extbase\Mvc\Controller\Argument::validate()`

* :php:`TYPO3\CMS\Extbase\Mvc\Controller\Arguments::validate()`

.. index:: PHP-API, FullyScanned, ext:extbase

tomasnorre:
I could take a look at this too. When the first one is finished.

sabbelasichon:
For this one i think you can use a so called configurable Rector: https://github.com/rectorphp/rector/blob/master/docs/rector_rules_overview.md#renamemethodrector

sabbelasichon:
No need for stubs and tests. The only thing that´s left is how to communicate that typo3-rector is doing this change for you. Because the configurable rectors are not part of the documentation.

PR #1626, opened 2020-11-21
[TASK] Change default caching framework names (#1253)

tomasnorre:
Resolves: #1253

### DIFF
--- a/src/Rector/v10/v0/ChangeDefaultCachingFrameworkNamesRector.php
+++ b/src/Rector/v10/v0/ChangeDefaultCachingFrameworkNamesRector.php
@@ -9,7 +9,7 @@ use PhpParser\Node\Expr\MethodCall;
 use Rector\Core\Rector\AbstractRector;
 use Rector\Core\RectorDefinition\CodeSample;
 use Rector\Core\RectorDefinition\RectorDefinition;
-use TYPO3\CMS\Core\Core\CacheManager;
+use TYPO3\CMS\Core\Cache\CacheManager;
 
 /**
  * @see https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.0/Deprecation-88366-DefaultCachingFrameworkCacheNamesChanged.html
--- a/src/Rector/v10/v0/ChangeDefaultCachingFrameworkNamesRector.php
+++ b/src/Rector/v10/v0/ChangeDefaultCachingFrameworkNamesRector.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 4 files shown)


tomasnorre:
I was a little eager to move on, so I continued with this one. 
I expect the feedback from #1059 will apply to this one too, but will see. :) 

This one isn't working yet though. I have commented on the code where I have problems. But think it makes more sense to wait for the feedback.

src/Rector/v10/v0/ChangeDefaultCachingFrameworkNamesRector.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        return [MethodCall::class];
+    }
+
+    /**
+     * @param MethodCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->isMethodStaticCallOrClassMethodObjectType($node, CacheManager::class)) {
```
tomasnorre, line 32: This check isn't working like I expected, always hit this one, If I remove this check and the 
```php 
if (! $this->isName($node->name, 'getCache')) {
    return null;
}
``` 

Then I can see that the change of the argument is working as expected. So not that far from something working. But there might be a simpler rule for this one too as mentioned in the comment on the PR itself.

sabbelasichon: @tomasnorre The things is this line here use TYPO3\CMS\Core\Core\CacheManager; Remove one of the core subprefix and it should work

src/Rector/v10/v0/ChangeDefaultCachingFrameworkNamesRector.php
```diff
@@ -0,0 +1,75 @@
...
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\Rector\v10\v0;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\MethodCall;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\RectorDefinition\CodeSample;
+use Rector\Core\RectorDefinition\RectorDefinition;
+use TYPO3\CMS\Core\Core\CacheManager;
```
sabbelasichon, line 12: This is wrong. Change Core with Cache and it should work

tomasnorre: I overlooked that.

sabbelasichon:
Very cool. One last thing: Could you please add the rector to the appropriate config file too?

tomasnorre:
>Very cool. One last thing: Could you please add the rector to the appropriate config file too? 

Done, but that I didn't do for the first rector that you have already merged.

PR #1633, opened 2020-11-22
[TASK] Remove CharsetConvert Parameters (#503)

### LINKED ISSUE #503: Breaking: #74031 - CharsetConverter parameters removed
Breaking: #74031 - CharsetConverter parameters removed

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.0/Breaking-74031-CharsetConverterParametersRemoved.html

======================================================

Breaking: #74031 - CharsetConverter parameters removed

======================================================

See :issue:`74031`

Description

===========

The second parameter for the method `CharsetConverter->entities_to_utf8()` has been removed.

The second and the third parameter of `CharsetConverter->utf8_to_numberarray()` has been removed.

Impact

======

When calling `CharsetConverter->entities_to_utf8()`, string-HTML entities (like &amp; or &pound;) will be converted

to UTF-8 as well at all times. Previously this behaviour was configurable.

When calling `CharsetConverter->utf8_to_numberarray()`, string-HTML entities (like

&amp; or &pound;) will be converted to UTF-8 as well at all times. Additionally instead

of integer numbers the real UTF-8 char is returned at any times. Previously these

behaviours were configurable.

Affected Installations

======================

Installations with custom extensions that used these methods directly in PHP code.

Migration

=========

Remove these parameters from the calling PHP code.

.. index:: PHP-API

tomasnorre:
Resolves #503

### DIFF
--- a/src/Rector/v8/v0/RemoveCharsetConverterParametersRector.php
+++ b/src/Rector/v8/v0/RemoveCharsetConverterParametersRector.php
@@ -37,15 +37,7 @@ final class RemoveCharsetConverterParametersRector extends AbstractRector
             return null;
         }
 
-        if (count($node->args) <= 1) {
-            return null;
-        }
-
-        if (3 === count($node->args)) {
-            $this->removeNode($node->args[2]);
-        }
-
-        $this->removeNode($node->args[1]);
+        $node->args = [$node->args[0]];
 
         return $node;
     }
--- a/src/Rector/v8/v0/RemoveCharsetConverterParametersRector.php
+++ b/src/Rector/v8/v0/RemoveCharsetConverterParametersRector.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\Rector\v8\v0;
+
+use PhpParser\Node;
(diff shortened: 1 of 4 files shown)


src/Rector/v8/v0/RemoveCharsetConverterParametersRector.php
```diff
@@ -0,0 +1,75 @@
...
+
+        if (! $this->isNames($node->name, ['entities_to_utf8', 'utf8_to_numberarray'])) {
+            return null;
+        }
+
+        if (count($node->args) <= 1) {
+            return null;
+        }
+
+        if (3 === count($node->args)) {
```
tomasnorre, line 44: I didn't see a better way to do this, if checking the function-name I would have to check the arguments count anyway to avoid the risk of undefined index.

sabbelasichon: @tomasnorre Just by reading i think it should be enough to set $node->args = [$node->args[0]];
 
It is not necessary to remove the arguments explicitly. But you should return $node and not null. Because you modified it and subsequent rectors should be informed by it.

Issue #762, opened 2019-11-04
Deprecation: #80513 - DataHandler: Various methods and method arguments



sabbelasichon:
Deprecation: #80513 - DataHandler: Various methods and method arguments

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.7/Deprecation-80513-DataHandlerVariousMethodsAndMethodArguments.html

.. include:: ../../Includes.txt

=======================================================================

Deprecation: #80513 - DataHandler: Various methods and method arguments

=======================================================================

See :issue:`80513`

Description

===========

The method :php:`DataHandler->destPathFromUploadFolder()` has been marked as deprecated.

The fourth parameter :php:`$func` of the method :php:`DataHandler->extFileFunctions()` has been deprecated.

Impact

======

Calling the method :php:`DataHandler->destPathFromUploadFolder()` will trigger a deprecation log entry.

Calling the method :php:`DataHandler->extFileFunctions()` with the fourth parameter (usually set to

'deleteAll') will trigger a deprecation log entry.

Affected Installations

======================

Any installation with custom extension logic using the DataHandler and specifically these methods.

Migration

=========

Replace the function call :php:`DataHandler->destPathFromUploadFolder()` by prepend the

constant :php:`PATH_site` before the string to be handed over to the deprecated method.

Remove the fourth parameter of the callee of :php:`DataHandler->extFileFunctions()`.

.. index:: PHP-API, Backend

tomasnorre:
Will give this a spin

sabbelasichon:
Amazing what you have done so far. Love it.

PR #1637, opened 2020-11-22
[TASK] DataHandler Various methods and method arguments (#762)

tomasnorre:
Resolves: #762

### DIFF
--- a/src/Rector/v8/v7/DataHandlerVariousMethodsAndMethodArgumentsRector.php
+++ b/src/Rector/v8/v7/DataHandlerVariousMethodsAndMethodArgumentsRector.php
@@ -1,0 +1,75 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\Rector\v8\v7;
+
+use PhpParser\Node;
+use PhpParser\Node\Arg;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\StaticCall;
+use Rector\Core\Rector\AbstractRector;
+use Rector\Core\RectorDefinition\CodeSample;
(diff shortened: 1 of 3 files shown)


src/Rector/v8/v7/DataHandlerVariousMethodsAndMethodArgumentsRector.php
```diff
@@ -0,0 +1,75 @@
...
+    {
+        if (! $this->isMethodStaticCallOrClassMethodObjectType($node, DataHandler::class)) {
+            return null;
+        }
+
+        if ($this->isName($node->name, 'destPathFromUploadFolder')) {
+            ///** @var Arg[] $args */
+            //$args = $node->args;
+            //$firstArgument = array_shift($args);
+            //$dest = PATH_SITE . $firstArgument
```
tomasnorre, line 42: Here I'm a struggling again similar to the `echo` earlier. 

I have found out how to get the value of the argument given to the original function `destPathFromUploadFolder`, but don't find how to have the constant concatenated before and having the `$dest = ` kept. 

Please let me know if I am asking too much. This AST is all quite new to me. And if I don't find examples where I can peek into, I find it a little hard to grasp be honest.

sabbelasichon: You can ask as many questions as you like until i explained everything to you ;-)

So also here you have to use the node type \PhpParser\Node\Expr\BinaryOp\Concat and also the node type ConstFetch and Name. So in total it look like that:
```php
return new \PhpParser\Node\Expr\BinaryOp\Concat(new \PhpParser\Node\Expr\ConstFetch(new \PhpParser\Node\Name('PATH_site')),$firstArgument->value);
```

Issue #1000, opened 2019-11-04
Deprecation: #83883 - Page Not Found And Error handling in Frontend



sabbelasichon:
Deprecation: #83883 - Page Not Found And Error handling in Frontend

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.2/Deprecation-83883-PageNotFoundAndErrorHandlingInFrontend.html

.. include:: ../../Includes.txt

===================================================================

Deprecation: #83883 - Page Not Found And Error handling in Frontend

===================================================================

See :issue:`83883`

Description

===========

The following methods have been marked as deprecated:

* php:`TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController->pageUnavailableAndExit()`

* php:`TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController->pageNotFoundAndExit()`

* php:`TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController->checkPageUnavailableHandler()`

* php:`TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController->pageUnavailableHandler()`

* php:`TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController->pageNotFoundHandler()`

* php:`TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController->pageErrorHandler()`

These methods have been commonly used by third-party extensions to show that a page is not found,

a page is unavailable due to misconfiguration or the access to a page was denied.

Impact

======

Calling any of the methods above will trigger a deprecation warning.

Affected Installations

======================

Any installation with third-party PHP extension code calling these methods.

Migration

=========

Use the new `ErrorController` with its custom actions `unavailableAction()`, `pageNotFoundAction()` and

`accessDeniedAction()`.

Instead of exiting the currently running script, a proposed PSR-7-compliant response is returned which can be

handled by the third-party extension to enrich, return or customize exiting the script.

.. index:: Frontend, PHP-API, FullyScanned

ayacoo:
On it

sabbelasichon:
i.e.
$message = 'No entry found!';
$GLOBALS['TSFE']->pageNotFoundAndExit('No entry found.');
$response = GeneralUtility::makeInstance(ErrorController::class)->pageNotFoundAction(
                        $GLOBALS['TYPO3_REQUEST'],
                        $message
                    );
                    throw new ImmediateResponseException($response, 1590468229);

PR #1641, opened 2020-11-22
[WIP][TASK] Page Not Found And Error handling in Frontend

ayacoo:
Related #1000

First version to discuss:
1.) Is the construction basically ok? I would still outsource the individual functions to my own methods. Here is just one example, the others are similar.
2.) The test is green, but the initial line for the TypoScriptFrontendController unfortunately remains. Do you have any ideas?


sabbelasichon:
Closed in favour of #1699

sabbelasichon:
@ayacoo I had to dive a little bit deeper to understand what to do. Not every case is covered. But i would say that most of the common use cases are covered.

Issue #1635, opened 2020-11-22
Document all configurable rectors


sabbelasichon:

tomasnorre:
How should this be done? Do you have an example @sabbelasichon ?

sabbelasichon:
For now, i have no other idea than just create a .md File underneath the docs folder manually and put all the configured rectors into it. For example this one here: https://github.com/sabbelasichon/typo3-rector/blob/master/config/v8/typo3-80.php#L105.

tomasnorre:
Is it worth investing some time in a way to see how it's can be done automatically?

tomasnorre:
I haven't got it to run within the TYPO3 Rector due to dependency conflicts so could not test, 
but sounds like this could help us. 

https://github.com/symplify/rule-doc-generator

sabbelasichon:
@tomasnorre Looks promising. I think Tomas extracted this as a new repository. So, let us wait some time. If we upgrade to a new rector version where this package is included we will give it a try. So this ticket is on hold for now.

sabbelasichon:
I think the configurable rectors are not taking into account by the auto generation of the docs. I think it is too much work and not really worth it. We have already a lot of rectors to show. I am gonna close this issue.

PR #1650, opened 2020-11-24
[TASK] Replace JStop with addCssFiles in FormResultCompiler (#624)

### LINKED ISSUE #624: Deprecation: #75363 - Deprecate FormResultCompiler->JStop()
Deprecation: #75363 - Deprecate FormResultCompiler->JStop()

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.4/Deprecation-75363-DeprecateFormResultCompilerJStop.html

===========================================================

Deprecation: #75363 - Deprecate FormResultCompiler->JStop()

===========================================================

See :issue:`75363`

Description

===========

The method `JStop()` has been renamed to `addCssFiles()`. The old method name is

still present as a deprecated alias, which will be removed in TYPO3 v9.

Keep in mind that the method reads "JS top", not "J Stop".

Impact

======

Calling `JStop()` method will trigger a deprecation log entry.

Affected Installations

======================

Any extensions calling `JStop()`.

Migration

=========

Instead of :php:`JStop()` use :php:`addCssFiles()`.

.. index:: PHP-API, Backend

tomasnorre:
Resolves #624

### DIFF
--- a/config/v8/typo3-84.php
+++ b/config/v8/typo3-84.php
@@ -1,0 +1,20 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Renaming\Rector\MethodCall\RenameMethodRector;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use TYPO3\CMS\Backend\Routing\FormResultCompiler;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $containerConfigurator->import(__DIR__ . '/../services.php');
+    $services = $containerConfigurator->services();
+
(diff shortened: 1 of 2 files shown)


tomasnorre:
How to handle this? 

> The only thing that´s left is how to communicate that typo3-rector is doing this change for you. Because the configurable rectors are not part of the auto generated documentation.

That am still not sure off, but find in a little "unsatisfying" that there are still todos open when I pushed.

sabbelasichon:
@tomasnorre That´s all fine. I know that feeling. The comment is just something to keep in mind. I will try to come up with something. The issue is that one: https://github.com/sabbelasichon/typo3-rector/issues/1635

Issue #474, opened 2019-11-04
Deprecation: #73482 - $LANG->csConvObj and $LANG->parserFactory



sabbelasichon:
Deprecation: #73482 - $LANG->csConvObj and $LANG->parserFactory

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.0/Deprecation-73482-LANG-csConvObjAndLANG-parserFactory.html

.. include:: ../../Includes.txt

===============================================================

Deprecation: #73482 - $LANG->csConvObj and $LANG->parserFactory

===============================================================

See :issue:`73482`

Description

===========

The properties of LanguageService (also known as `$GLOBALS[LANG]`) csConvObj and parserFactory

have been marked as deprecated. Since these three PHP classes are not dependent on each other, they

can be decoupled. The getter method `getParserFactory()` has thus been marked as deprecated as well.

Impact

======

These properties will be removed in TYPO3 v9. Calling `LanguageService->getParserFactory()` will trigger a

deprecation log entry.

Affected Installations

======================

Installations with custom extension accessing the LanguageService properties and method above.

Migration

=========

Instantiate CharsetConverter (csConvObj) and LocalizationFactory (parserFactory) via `GeneralUtility::makeInstance`

directly if needed, as they are Singleton objects and then fetched from the General Utility Object container

functionalities.

.. index:: PHP-API

PR #1651, opened 2020-11-24
[TASK] Remove CsConvObj and ParserFactory from LanguageService::class (#474)

tomasnorre:
Resolves: #474

### DIFF
--- a/src/Rector/v8/v0/RemoveLangCsConvObjAndParserFactoryRector.php
+++ b/src/Rector/v8/v0/RemoveLangCsConvObjAndParserFactoryRector.php
@@ -16,7 +16,6 @@ use TYPO3\CMS\Core\Charset\CharsetConverter;
 use TYPO3\CMS\Core\Localization\LanguageService;
 use TYPO3\CMS\Core\Localization\LocalizationFactory;
 use TYPO3\CMS\Core\Utility\GeneralUtility;
-use TYPO3\CMS\Frontend\Plugin\AbstractPlugin;
 
 /**
  * @see https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.0/Deprecation-73482-LANG-csConvObjAndLANG-parserFactory.html
@@ -54,7 +53,7 @@ final class RemoveLangCsConvObjAndParserFactoryRector extends AbstractRector
             return $this->refactorLanguageServiceCall($node);
         }
 
-        return $this->refactorAbstractPluginCall($node);
+        return null;
     }
 
     /**
@@ -92,12 +91,7 @@ PHP
         if ($this->isLanguageServiceCall($node)) {
             return false;
         }
-
-        if ($node instanceof PropertyFetch) {
-            return true;
-        }
-
-        return ! $this->isMethodStaticCallOrClassMethodObjectType($node, AbstractPlugin::class);
+        return $node instanceof PropertyFetch;
     }
 
     private function isLanguageServiceCall(Node $node): bool
@@ -110,7 +104,7 @@ PHP
             return true;
         }
 
-        if ($this->typo3NodeResolver->isPropertyFetchOnAnyPropertyOfGlobals($node, 'LANG')) {
+        if ($this->typo3NodeResolver->isPropertyFetchOnAnyPropertyOfGlobals($node, Typo3NodeResolver::LANG)) {
             return true;
         }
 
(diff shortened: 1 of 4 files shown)


src/Rector/v8/v0/RemoveLangCsConvObjAndParserFactoryRector.php
```diff
@@ -0,0 +1,76 @@
...
+    /**
+     * @param MethodCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->isMethodStaticCallOrClassMethodObjectType($node, LanguageService::class)) {
+            return null;
+        }
+
+        if ($this->isName($node->name, 'csConvObj')) {
```
tomasnorre, line 39: I have doing it wrong here. I have changed the function call of `csConvObj()`  and not the property usage.
I'll look into this again.

sabbelasichon: Right. And we have to deal with the $GLOBALS['LANG'] here too. So you need the Typo3NodeResolver class and check if we are dealing with a propertyFetch on $GLOBALS. See i.e. https://github.com/sabbelasichon/typo3-rector/blob/master/src/Rector/v9/v5/RefactorPropertiesOfTypoScriptFrontendControllerRector.php#L43

tomasnorre:
This could be an inspiration https://github.com/sabbelasichon/typo3-rector/pull/1685

tomasnorre:
I have the rector working now with both languageService and `$GLOBALS['LANG'], but still fighting with some PHPStan feedback,

src/Rector/v8/v0/RemoveLangCsConvObjAndParserFactoryRector.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        if ($this->isLanguageServiceCall($node)) {
+            return false;
+        }
+
+        if ($node instanceof PropertyFetch) {
+            return true;
+        }
+
+        return ! $this->isMethodStaticCallOrClassMethodObjectType($node, AbstractPlugin::class);
```
sabbelasichon, line 100: This is unrelated to this Rector. Maybe some relict from the other one.

src/Rector/v8/v0/RemoveLangCsConvObjAndParserFactoryRector.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        if (! (property_exists($node, 'var') && null !== $node->var)) {
+            return false;
+        }
+
+        if ($this->isObjectType($node->var, LanguageService::class)) {
+            return true;
+        }
+
+        if ($this->typo3NodeResolver->isPropertyFetchOnAnyPropertyOfGlobals($node, 'LANG')) {
```
sabbelasichon, line 113: You can use the constants here too. Typo3NodeResolver::LANG

src/Rector/v8/v0/RemoveLangCsConvObjAndParserFactoryRector.php
```diff
@@ -0,0 +1,149 @@
...
+    {
+        if (! (property_exists($node, 'name') && null !== $node->name)) {
+            return null;
+        }
+
+        if (null === $this->getName($node->name)) {
+            return null;
+        }
+
+        if ('csConvObj' === $this->getName($node->name)) {
```
sabbelasichon, line 130: You could use isName here too for better readibility.

sabbelasichon (commented):
Thanks a lot. Besides some minor issues, looks good to me

src/Rector/v8/v0/RemoveLangCsConvObjAndParserFactoryRector.php
tomasnorre: How would that look like in this case. Cause if doing 

```php
 if (Typo3NodeResolver::LANG) {
            return true;
}
``` 

It's always true and we never this the last return statement

sabbelasichon: No, i mean instead of the String LANG here. The call is right.

tomasnorre: I see.

PR #1685, opened 2020-11-27
Issue 536

sabbelasichon:


src/Rector/v8/v2/UseHtmlSpecialCharsDirectlyForTranslationRector.php
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    /**
+     * @codeCoverageIgnore
+     */
+    public function getDefinition(): RectorDefinition
+    {
+        return new RectorDefinition('htmlspecialchars directly to properly escape the content.', [
+            new CodeSample(<<<'PHP'
+PHP
```
sabbelasichon, line 38: Need to add the CodeExamples here

tomasnorre (commented):
From reading, itlooks good to me. 
 
The missing path of example code you have pointed out yourself already.

src/Rector/v8/v2/UseHtmlSpecialCharsDirectlyForTranslationRector.php
```diff
@@ -0,0 +1,153 @@
...
+     */
+    public function getNodeTypes(): array
+    {
+        return [MethodCall::class];
+    }
+
+    /**
+     * @param MethodCall $node
+     */
+    public function refactor(Node $node): ?Node
```
tomasnorre, line 80: I'm trying to finish the https://github.com/sabbelasichon/typo3-rector/pull/1651, and looking into this as you suggested.

I don't get my head around this code though.

The `rector()`-function calls `$this->shouldSkip()`. `$this->shouldSkip()` calls the `$this->isLanguageServiceCall()`. But the`rector()`-function itself also calls the `$this->isLanguageServiceCall()` and that I don't get.

What is the difference when calling it from the `$this->shouldSkip()` and the `rector()`-function ?

sabbelasichon: The shouldSkip function does also a check if it is a call from an object of class AbstractPlugin.

Issue #1412, opened 2020-05-06
Deprecation: #89870 - New PSR-14 Events for Extbase-related signals



sabbelasichon:
Deprecation: #89870 - New PSR-14 Events for Extbase-related signals

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/10.3/Deprecation-89870-NewPSR-14EventsForExtbase-relatedSignals.html

.. include:: ../../Includes.txt

===================================================================

Deprecation: #89870 - New PSR-14 Events for Extbase-related signals

===================================================================

See :issue:`89870`

Description

===========

The following signals have been marked as deprecated in favor of new PSR-14 events:

- :php:`TYPO3\CMS\Extbase\Mvc\Dispatcher::afterRequestDispatch`

- :php:`TYPO3\CMS\Extbase\Mvc\Controller\ActionController::beforeCallActionMethod`

- :php:`TYPO3\CMS\Extbase\Persistence\Generic\Mapper\DataMapper::afterMappingSingleRow`

- :php:`TYPO3\CMS\Extbase\Persistence\Generic\Backend::beforeGettingObjectData`

- :php:`TYPO3\CMS\Extbase\Persistence\Generic\Backend::afterGettingObjectData`

- :php:`TYPO3\CMS\Extbase\Persistence\Generic\Backend::endInsertObject`

- :php:`TYPO3\CMS\Extbase\Persistence\Generic\Backend::afterUpdateObject`

- :php:`TYPO3\CMS\Extbase\Persistence\Generic\Backend::afterPersistObject`

- :php:`TYPO3\CMS\Extbase\Persistence\Generic\Backend::afterRemoveObject`

The method :php:`emitBeforeCallActionMethodSignal` in :php:`ActionController`

has been marked as deprecated and is not called by Extbase itself anymore.

Impact

======

Using any of the signals will still work as expected, but will trigger

a PHP :php:`E_USER_DEPRECATED` error.

Calling the method :php:`emitBeforeCallActionMethodSignal` will trigger a

PHP :php:`E_USER_DEPRECATED` error.

Affected Installations

======================

TYPO3 installations with extensions using the Extbase framework and

Extbase-internal hooks.

Migration

=========

The following new PSR-14-based Events should be used instead:

- :php:`TYPO3\CMS\Extbase\Event\Mvc\AfterRequestDispatchedEvent`

[...]

PR #1693, opened 2020-11-29
[WIP][TASK] Use PSR-14 Events instead of Signals (#1412)

tomasnorre:


src/Rector/v10/v3/UsePSR14EventsExtbaseRelatedSignalsRector.php
```diff
@@ -0,0 +1,86 @@
...
+                return null;
+            }
+
+            $argumentValue = $this->getValue($firstArgument->value);
+
+            if (null === $argumentValue) {
+                return null;
+            }
+
+            return $this->createMethodCall($this->createPropertyFetch($node->var, 'eventDispatcher'), 'dispatch',
```
tomasnorre, line 60: Hi @sabbelasichon,

It's still work in progress, but I'm currently stucked here. Hope you can give me some pointers. As you see I end up with a `$new->BeforeActionCallEvent` instead of `new BeforeActionCallEvent` + the `static::class` shouldn't be a string.

I'm pretty sure it's fairly easy to fix, but I don't find the function needed for this. 

```php
     public function method(): void
     {
         $actionController = GeneralUtility::makeInstance(ActionController::class);
-        $actionController->eventDispatcher->dispatch(new BeforeActionCallEvent(static::class, 'indexAction', []));
+        $actionController->eventDispatcher->dispatch($new->BeforeActionCallEvent('static::class', 'indexAction', []));
     }
```

tomasnorre: Currently playing around with the `New_`

tomasnorre: This is my best shot so far. 

``` 
$this->createFuncCall( 'new BeforeActionCallEvent', ['static::class', 'indexAction', $firstArgument]),
```
But I'm pretty sure it's not correct with the "new BeforeActionCallEvent" as I'm still missing the `use statement` the `static::class` is still WIP. Didn't find a solution for that yet either.

sabbelasichon:
@tomasnorre I think that we do not have to change the caller side but rather the callee one (the listener). So if someone is connecting to the old slot, it would be awesome if you rewrite this to the new PSR-14 Event System. But i think, this is really tough to do. I have to think about it.

sabbelasichon:
I came to the conclusion that this is too much for now. Close this issue for now.

tomasnorre:
Issue is closed so PR will be dropped.

PR #1861, opened 2021-01-06
Fix configured path to typo3.constants.php

DanielSiepmann:
The configured path to typo3.constants.php file is wrong.
It is now changed to expected location.
This already fixes the following issue for default setups.
Still if packages are installed somewhere else, the path would still be
wrong.

[ERROR] Could not process "*" file, due to:
        "File or directory 'typo3.constants.php' not found.".

### DIFF
--- a/templates/rector.php.dist
+++ b/templates/rector.php.dist
@@ -48,9 +48,9 @@ return static function (ContainerConfigurator $containerConfigurator): void {
 
     // If you have trouble that rector cannot run because some TYPO3 constants are not defined add an additional constants file
     // Have a look at https://github.com/sabbelasichon/typo3-rector/blob/master/typo3.constants.php
-    $parameters->set(Option::AUTOLOAD_PATHS, [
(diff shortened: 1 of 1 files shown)


DanielSiepmann:
I know this is not a real solution, but should already improve current situation.

I'm wondering why no one else receives the issue?
Maybe I did something wrong, but I just followed the docs to install, generate config, and execute.

sabbelasichon:
@DanielSiepmann Actually, it is not a bug it is a feature. I should have better commented out this. The purpose is to copy over this file manually, because in most of the cases you don´t need it. But in some setups it is needed. The comment was a bit misleading. I think i am gonna copy it over with the typo3-init command. Because your solution does not work for the phar file and could lead to other problems.

DanielSiepmann:
I've updated with the simple solution you've mentioned, just keep this configuration commented out.
The comment already explains the situation, so let's use sane defaults.

PR #1864, opened 2021-01-06
Prevent possible duplicate constant definition

DanielSiepmann:
Under some circumstances it might happen that TYPO3_COMPOSER_MODE is
already defined.
This change fixes the following notice:

Notice: Constant TYPO3_COMPOSER_MODE already defined in
/vendor/ssch/typo3-rector/typo3.constants.php on line 42


DanielSiepmann:
I don't know the circumstances, but we can investigate on my setup if necessary.

PR #1941, opened 2021-02-04
[BUGFIX] Preserve formatting of TCA when no changes are necessary

MK-42:
if no changes to the syntax tree are necessary, we should not alter it, to preserve the code formatting.

this reduces noise in rector output and git logs


MK-42:
thanks a lot!

PR #1944, opened 2021-02-05
[BUGFIX] Return $node if-and-only-if the ast has been changed. Otherw…

### LINKED ISSUE #1942: Rectors must only return $node if they changed the AST
according to
https://github.com/rectorphp/rector/blob/0a3dc08d61621d44db4324a5385b79a565414f39/docs/create_own_rule.md

rectors must return null if no AST change has been made. The current code leads to many false-positives where rectors are listed as they changed the code, but nothing changed at all

MK-42:
…ise return null instead

fixes #1942

This is a quick turnover of all rectors. I'm trusting the test here, could not review every single change in detail. Feel free to pick this in chunks or something

### DIFF
--- a/src/Rector/v7/v4/DropAdditionalPaletteRector.php
+++ b/src/Rector/v7/v4/DropAdditionalPaletteRector.php
@@ -57,15 +57,16 @@ final class DropAdditionalPaletteRector extends AbstractRector
         $types = $this->extractTypes($node);
 
         if (! $types instanceof ArrayItem) {
-            return $node;
+            return null;
         }
 
         $typesItems = $types->value;
 
         if (! $typesItems instanceof Array_) {
-            return $node;
+            return null;
         }
 
+        $hasAstBeenChanged = false;
         foreach ($typesItems->items as $typesItem) {
             if (! $typesItem instanceof ArrayItem) {
                 continue;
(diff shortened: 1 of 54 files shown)


src/Rector/v7/v4/DropAdditionalPaletteRector.php
```diff
@@ -129,10 +130,14 @@ public function refactor(Node $node): ?Node
                     continue;
                 }
                 $typeItem->value = new String_(implode(',', $newFieldStrings));
+                $hasAstBeenChanged = true;
             }
         }
 
-        return $node;
+        if ($hasAstBeenChanged) {
```
sabbelasichon, line 137: Could you please change this to `return $hasAstBeenChanged ? $node : null;` to be consistent.

sabbelasichon (commented):
Thanks. Could you please change this 
`if ($hasAstBeenChanged) {
 return $node;
}
return null

src/Rector/v7/v4/DropAdditionalPaletteRector.php
MK-42: of course. Just realized while going through the files that a ternary would be nice and didn't go back to update the rest

PR #1962, opened 2021-02-10
[Testing] Make use of config new Community config-based test case

TomasVotruba:
Follow up to https://github.com/sabbelasichon/typo3-rector/pull/1961 - waits on the merge now

Closes https://github.com/rectorphp/rector/issues/5478

PR #1964, opened 2021-02-11
[CI] Improve workflows

TomasVotruba:
Last follow up to https://github.com/sabbelasichon/typo3-rector/pull/1962

Tip for automated weekly coverage: https://github.com/rectorphp/rector/blob/master/.github/workflows/weekly_code_coverage.yaml

<br>

Btw, what do you use coverage in practise? I'm thinking about it lately and I never work with the coverage number in any way :)

### DIFF
--- a/.github/workflows/code_coverage.yaml
+++ b/.github/workflows/code_coverage.yaml
@@ -1,7 +1,6 @@
 name: Code Coverage
 
 on:
-    pull_request: null
     push:
         branches:
             - master
--- a/.github/workflows/code_coverage.yaml
+++ b/.github/workflows/code_coverage.yaml
@@ -11,14 +11,7 @@ jobs:
         runs-on: ubuntu-latest
         steps:
             -   uses: actions/checkout@v2.0.0
-            -   id: composer-cache
-                run: echo "::set-output name=dir::$(composer config cache-files-dir)"
-            -   uses: actions/cache@v2
-                with:
-                    path: ${{ steps.composer-cache.outputs.dir }}
-                    key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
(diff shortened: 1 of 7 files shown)


.github/workflows/tests.yaml
```diff
@@ -11,25 +11,20 @@ jobs:
...
-            -   uses: actions/cache@v1
-                with:
-                    path: ${{ steps.composer-cache.outputs.dir }}
-                    key: ${{ runner.os }}-composer-${{ hashFiles('**/composer.lock') }}
-                    restore-keys: |
-                        ${{ runner.os }}-composer-
+
             -   uses: shivammathur/setup-php@v2
                 with:
-                    php-version: 7.3
```
TomasVotruba, line 29: It seems tests only run on PHP 7.3.
That explain why I had so many errors on PHP 8.0 :)

.github/workflows/code_coverage.yaml
```diff
@@ -11,22 +10,16 @@ jobs:
...
             -   uses: shivammathur/setup-php@v2
                 with:
                     php-version: 7.3
                     coverage: none
                 env:
                     COMPOSER_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-            -   run: composer update --no-progress
+            # see https://github.com/ramsey/composer-install
+            -   uses: "ramsey/composer-install@v1"
```
TomasVotruba, line 22: Little cool trick to simplify cached composer install

TomasVotruba:
Ready to review :heavy_check_mark:

.github/workflows/tests.yaml
sabbelasichon: @TomasVotruba Oh no, stupid me. Thanks for fixing.

PR #1970, opened 2021-02-15
[FEATURE] Migrate ObjectManager->get() to GeneralUtility::makeInstance()

DanielSiepmann:
This migration assumes developers already migrated to dependency
injection where possible and useful.
All existing calls will be migrated.
In order to have proper working code, a Services.yaml is necessary.

Relates: #1883

### DIFF
--- a/config/v10/typo3-104.php
+++ b/config/v10/typo3-104.php
@@ -4,6 +4,8 @@ declare(strict_types=1);
 
 use Rector\Renaming\Rector\StaticCall\RenameStaticMethodRector;
 use Rector\Renaming\ValueObject\RenameStaticMethod;
+use Rector\Transform\Rector\MethodCall\MethodCallToStaticCallRector;
+use Rector\Transform\ValueObject\MethodCallToStaticCall;
 use Ssch\TYPO3Rector\Rector\Migrations\RenameClassMapAliasRector;
 use Ssch\TYPO3Rector\Rector\v10\v4\SubstituteGeneralUtilityMethodsWithNativePhpFunctionsRector;
 use Ssch\TYPO3Rector\Rector\v10\v4\UnifiedFileNameValidatorRector;
@@ -42,4 +44,16 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             ],
         ]]);
     $services->set(UseFileGetContentsForGetUrlRector::class);
+    $services->set('typo3_objectmanagerget_to_generalutilitymakeinstance')
+        ->class(MethodCallToStaticCallRector::class)
+        ->call('configure', [[
+            MethodCallToStaticCallRector::METHOD_CALLS_TO_STATIC_CALLS => ValueObjectInliner::inline([
+                new MethodCallToStaticCall(
+                    \TYPO3\CMS\Extbase\Object\ObjectManagerInterface::class,
+                    'get',
+                    \TYPO3\CMS\Core\Utility\GeneralUtility::class,
+                    'makeInstance'
+                ),
+            ]),
+        ]]);
 };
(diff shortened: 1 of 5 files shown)


config/v10/typo3-104.php
```diff
@@ -35,4 +36,16 @@
         ],
     ]]);
     $services->set(UseFileGetContentsForGetUrlRector::class);
+    $services->set('typo3_objectmanagerget_to_generalutilitymakeinstance')
```
DanielSiepmann, line 39: I'm not sure about a proper naming convention for the services. Maybe you should define a convention upfront, and I'll adjust the PR.

sabbelasichon (commented):
i would not add this rule in the v10 set. This is confusing and many people do not know how to react. Should be part of the experimental ones or somewhere else.

DanielSiepmann:
Sure, would be fine for me.

There is only one experimental rector yet, and this is a concrete rector.
This one will be configuration. Should I only add the configuration somewhere?
Or should we add some kind of experimental set / configuration? Even those experimental changes are for specific TYPO3 versions.

sabbelasichon:
@DanielSiepmann Good questions. I am not sure. Tiny change big impact. I would just add a experimental config at the beginning. Let´s see how far we will get. In the end, everyone can have a look at this configuration and select the desired rules. What do you think?

MK-42:
What I see a lot in real life projects is the legacy instantiation of extbase domain objects with object manager. These could be migrated without breaking most projects with the proposed replacement of the get() call.

Maybe that could be added to the normal sets, and the changes that require a services.yaml could be incorporated in an experimental set?

I haven't looked into the whole new DI concept in v10, maybe we can provide some default services.yaml or extend an existing one along with this rector? Is it feasible to write such a config with a rector?

MK-42:
On a second thought: objectManager->get for domain models is mostly used to be able to plug different implementations into the extbase context, or extending domain objects via inheritance.

For this to work there is the extbase containers ->Register ImplementationClassnallName, which must now be migrated to am XCLASS registration (if used). Maybe we could provide this as well? What do You think about this approach? This would of course only make sense for domain models.

sabbelasichon:
@MK-42 @DanielSiepmann  Do you  know what´s the proposed solution to achieve the old behaviour of initializeObject and injections of service into entities (i know it is a code smell, but it is possible).

tests/Rector/v10/v4/SubstituteObjectManagerGetWithGeneralUtilityMakeInstanceRector/config/configured_rule.php
```diff
@@ -0,0 +1,26 @@
...
+use Rector\Transform\Rector\MethodCall\MethodCallToStaticCallRector;
+use Rector\Transform\ValueObject\MethodCallToStaticCall;
+use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;
+use Symplify\SymfonyPhpConfig\ValueObjectInliner;
+
+return static function (ContainerConfigurator $containerConfigurator): void {
+    $containerConfigurator->import(__DIR__ . '/../../../../../../config/services.php');
+
+    $services = $containerConfigurator->services();
+    $services->set('typo3_objectmanagerget_to_generalutilitymakeinstance')
```
sabbelasichon, line 14: I am nitpicking here, but i would prefer typo3_objectmanager_get_to_generalutility_makeinstance

DanielSiepmann:
> @MK-42 @DanielSiepmann Do you know what´s the proposed solution to achieve the old behaviour of initializeObject and injections of service into entities (i know it is a code smell, but it is possible).

No, and I would guess there is none. As far as I know Alexander is still more or less alone on Extbase front and tries to remove unnecessary stuff. Speaking of DI and makeInstance, I would guess one should better initialize within the constructor.

MK-42:
> @MK-42 @DanielSiepmann Do you know what´s the proposed solution to achieve the old behaviour of initializeObject and injections of service into entities (i know it is a code smell, but it is possible).

I think using injections via methods or property injection is an antipattern and there is no migration path for that.
Maybe we could introduce a phpstan error via saschas phpstan package. This would at least show an error for people who use injections in projects.
As for initializeObject I think that the framework should call that method still. But I can't find a prove for that at the moment.

The only thing rector could do imho would be to refactor that to GeneralUtility::makeInstance and objectManager->get. But that would not really improve the code, I think.

Issue #1984, opened 2021-02-18
Add more extensions to composer sets


sabbelasichon:
Maybe it is useful to add also more extensions to the composer configs to ease the upgrade process. I am not sure if this is useful. It always depends on the preferences of the developer what version constraint to use. But maybe we could say, upgrade to the at least minimum version for the targeted TYPO3 Version, i.e.

1. https://packagist.org/packages/apache-solr-for-typo3/solr#10.0.0
2. https://packagist.org/packages/directmailteam/direct-mail#7.0.0
3. https://packagist.org/packages/in2code/powermail#8.3.0
4. https://packagist.org/packages/localizationteam/l10nmgr#10.0.1
5. https://packagist.org/packages/jigal/t3adminer#9.4.1
6. https://packagist.org/packages/yoast-seo-for-typo3/yoast_seo#7.1.3
7. https://packagist.org/packages/bk2k/bootstrap-package#11.0.2
8. https://packagist.org/packages/fluidtypo3/vhs
9. https://packagist.org/packages/mask/mask#v6.4.4

tomasnorre:
I'm not sure what is meant by this issue. Can you please elaborate a little more.

sabbelasichon:
@tomasnorre Sure. I have added some new sets to automatically update the composer.json in the root of the directory which rector runs against. I.e. https://github.com/sabbelasichon/typo3-rector/blob/master/config/typo3-104-composer-packages.php#L74. It would be convenient to add some popular extensions which are compatible with the targeted TYPO3 version to the sets.

tomasnorre:
Ah. I see. 

When would an extension classify as popular?

sabbelasichon:
This is up to us. In the end it can be every extension but i don´t want to add every possible extension there.

tomasnorre:
Wouldn't this be something can in Theory could be autogenerated from the packagist information? 
https://packagist.org/?type=typo3-cms-extension 

All the requirements' information is stored on packagist, it "just" needs to be parsed. 
I would make the option optinal though, as I don't always want to auto update extension in case there are breaking changes etc. in the extension, i might want to do it by hand. not all extension will have rector rules for updating :)

sabbelasichon:
Nice one. Yes, could be done automatically. It is not about updating the code of extension in your custom code. It is only about updating the composer.json to the least required version. For example, you cannot update to TYPO3 Version 10 without having the package aoepeople/crawler:^9.0 at least.

[...]

PR #2004, opened 2021-02-27
[TASK] Add composer packages

sabbelasichon:
Resolves: #1984


config/composer/typo3-104-composer-packages-extensions.php
```diff
@@ -0,0 +1,1089 @@
...
+        new PackageAndVersion('aoe/restler', '^0.7.0'),
+        new PackageAndVersion('aoe/sentry-client-js', '^0.4.0'),
+        new PackageAndVersion('aoe/staticpub', '^1.4.0'),
+        new PackageAndVersion('aoe/staticpub-pageexport', '^0.2.0'),
+        new PackageAndVersion('aoe/t3deploy-ext', '^1.4.0'),
+        new PackageAndVersion('aoe/truncate-job', '^0.2.1'),
+        new PackageAndVersion('aoe/update-refindex', '^0.2.5'),
+        new PackageAndVersion('aoe/varnish', '^0.0.1'),
+        new PackageAndVersion('aoepeople/cachemgm', '^2.5.0'),
+        new PackageAndVersion('aoepeople/crawler', '^6.4.1'),
```
tomasnorre, line 55: If I read this correct something is not correct with the parsing. You need ^9.0.0 of the aoepeople/crawler to have it working with TYPO 10.4

config/composer/typo3-110-composer-packages-extensions.php
```diff
@@ -0,0 +1,376 @@
+<?php
```
tomasnorre, line 1: Again if I read this correctly, this cannot be right, there isn't that many extensions compatible with TYPO3 11.0 yet.

utils/composer-packages/src/ComposerPackageParser.php
```diff
@@ -0,0 +1,78 @@
...
+use Ssch\TYPO3Rector\ComposerPackages\Collection\ExtensionCollection;
+use Ssch\TYPO3Rector\ComposerPackages\ValueObject\ComposerPackage;
+use Ssch\TYPO3Rector\ComposerPackages\ValueObject\ExtensionVersion;
+use Ssch\TYPO3Rector\ComposerPackages\ValueObject\Typo3Version;
+
+final class ComposerPackageParser implements PackageParser
+{
+    public function parsePackage(string $content, ComposerPackage $composerPackage): ExtensionCollection
+    {
+        $json = Json::decode($content, Json::FORCE_ARRAY);
```
tomasnorre, line 19: Isn't it enough with the default PHP json functions? Why pull a external dependency here?

utils/composer-packages/src/ValueObject/ExtensionVersion.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\ComposerPackages\ValueObject;
+
+use Rector\Composer\ValueObject\PackageAndVersion;
+use Stringable;
+use Webmozart\Assert\Assert;
```
tomasnorre, line 9: What is the difference from this to https://packagist.org/packages/beberlei/assert besides way more downloads on packagist

config/composer/typo3-110-composer-packages-extensions.php
sabbelasichon: True. The extension authors did something wrong which is difficult to catch. For example https://packagist.org/packages/dmk/mkpostman#1.0.3 is compatible regarding the composer.json definition. I have to find a better solution to prevent this. But didn´t come up with one yet.

utils/composer-packages/src/ComposerPackageParser.php
sabbelasichon: Yes, should be enough. But then i have to implement the unhappy path myself. If the json is not valid, i have to catch this and throw an exception. The method is just convenient.

utils/composer-packages/src/ValueObject/ExtensionVersion.php
sabbelasichon: I am using the webmozart/assert library since ages. Don´t know any real difference for the most common assertions.

PR #2048, opened 2021-03-12
[BUGFIX] Apply new navigation component id only to "typo3-pagetree"

### LINKED ISSUE #2047: "UseNewComponentIdForPageTreeRector" should only be applied to "typo3-pagetree" value
## Problem

With #1614, the migration for the new navigation component id for page tree was introduced. This migration applies whenever the configuration key `navigationComponentId` is set, no matter to which value. This leads to problems, f.e. when custom implementations are used or in case the navigation component should be disabled (this was possible by defining an empty value).

## Example

The following configuration was not affected by the deprecation:

```php
\TYPO3\CMS\Extbase\Utility\ExtensionUtility::registerModule(
    // ...
    [
        // ...
        'navigationComponentId' => ''
    ]
);
```

The introduced rector however rewrites the code as follows:

```php
\TYPO3\CMS\Extbase\Utility\ExtensionUtility::registerModule(
    // ...
    [
        // ...
        'navigationComponentId' => 'TYPO3/CMS/Backend/PageTree/PageTreeElement'
    ]
);
```

## Solution

The migration should only happen if the configuration value is explicitly set to `typo3-pagetree`. This should cover all cases of the introduced deprecation as written in the [changelog](https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.0/Deprecation-82426-Typo3-pagetreeNavigationComponentName.html):

> The name [of] the navigation component name `typo3-pagetree` has been marked as deprecated. `TYPO3/CMS/Backend/PageTree/PageTreeElement` should be used instead.

Example of an affected configuration:

```php
\TYPO3\CMS\Extbase\Utility\ExtensionUtility::registerModule(
    // ...
    [
        // ...
        'navigationComponentId' => 'typo3-pagetree'
    ]
);
```

eliashaeussler:
Resolves #2047

### DIFF
--- a/src/Rector/v9/v0/UseNewComponentIdForPageTreeRector.php
+++ b/src/Rector/v9/v0/UseNewComponentIdForPageTreeRector.php
@@ -64,6 +64,9 @@ final class UseNewComponentIdForPageTreeRector extends AbstractRector
             if ('navigationComponentId' !== $this->valueResolver->getValue($item->key)) {
                 continue;
             }
+            if ('typo3-pagetree' !== $this->valueResolver->getValue($item->value)) {
+                continue;
+            }
             $item->value = new String_('TYPO3/CMS/Backend/PageTree/PageTreeElement');
             $hasAstBeenChanged = true;
         }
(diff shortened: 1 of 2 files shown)


sabbelasichon:
Thanks. Good catch

PR #2099, opened 2021-03-24
Add recommendation regarding single TYPO3 extension

kitzberger:


sabbelasichon:
@kitzberger Thanks, could you also add this option in the default template: https://github.com/sabbelasichon/typo3-rector/blob/master/templates/rector.php.dist#L40

Issue #2100, opened 2021-03-24
v9 - TCA type group internal_type file and file_reference


helsner:
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/9.5/Deprecation-86406-TCATypeGroupInternal_typeFileAndFile_reference.html

With v9 the support for the internal_type=file has been dropped.
The following example can be rewritten just like below:
```

'feedback_image' => [
    'exclude' => 1,
    'label' => 'FeedbackImage',
    'config' => [
        'type' => 'group',
        'internal_type' => 'file',
        'allowed' => $GLOBALS['TYPO3_CONF_VARS']['GFX']['imagefile_ext'],
        'max_size' => '20000',
        'uploadfolder' => 'fileadmin/feedbacks',
        'maxitems' => '1',
    ],
],

Turns into:

'feedback_image' => [
    'exclude' => 1,
    'label' => 'FeedbackImage',
    'config' => \TYPO3\CMS\Core\Utility\ExtensionManagementUtility::getFileFieldTCAConfig(
        'feedback_image',
        [
            'max_size' => '20000',
            'uploadfolder' => 'fileadmin/feedbacks',
            'maxitems' => 1,
            'appearance' => [
                'createNewRelationLinkTitle' => 'LLL:EXT:frontend/Resources/Private/Language/locallang_ttc.xlf:images.addFileReference',
            ],
        ],
        $GLOBALS['TYPO3_CONF_VARS']['GFX']['imagefile_ext']
    ),
],
```

A slight conflict occurred though:
The files are handled differently with FAL which makes a migration necessary, which can be build up just like here
https://review.typo3.org/c/Packages/TYPO3.CMS/+/54830/8/typo3/sysext/install/Classes/Updates/BackendLayoutIconUpdateWizard.php#1
Also the FE probably needs adaption.

To me the last two mentioned points are definitely not the job of rector.
The question would be if we can use rector to help out the DEV at this point by marking the TCA to be checked manually.
I checked the TCA Migration option in the BE with my v10 and did not a notification about the outdated segment anymore - rector could come to rescue in cases like this

helsner:
Most importantly: this is a suggestion.
I pointed out the contra aspects in the end and would fully understand a rejection.

sabbelasichon:
[...]

PR #2113, opened 2021-03-29
[TASK] Refactor TCA type group internal_type file and file_reference

sabbelasichon:
Resolves: #2100


helsner (approved):
I checked within my real project and it worked perfectly fine!

PR #2152, opened 2021-04-09
do not build scoped version on PR

TomasVotruba:
This addresses previously mentioned issue, when scoped version is build on PR. It should happen on merge to master or release


sabbelasichon:
You are too fast. I was just sitting here and wanted to push. Thanks a lot.

Issue #2162, opened 2021-04-09
ReplaceExtKeyWithExtensionKeyRector misses out sys_template.php


helsner:
According to best practice the addStaticFile should be placed in sys_template.php as TCA override
https://docs.typo3.org/m/typo3/reference-coreapi/10.4/en-us/ExtensionArchitecture/ConfigurationFiles/Index.html#id3

The ReplaceExtKeyWithExtensionKeyRector responsible for this checks for the file and filters out everything besides ext_localconf.php && ext_tables.php

We could extend the FileHelperTrait to provide a function to check for sys_template and check this within the rector

Code example that is not transformed:
`\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addStaticFile($_EXTKEY, 'Configuration/TypoScript', 'Example');`

@sabbelasichon you can write/approve if that seems like a valid solution to you, i would provide it then

sabbelasichon:
Yes, sure. Would be useful. Go for it. If i can help you, let me know and we will have a slack call.

ayacoo:
I ran into this rector today and have the following initial situation

Configuration/TCA/Overrides/sys_template.php
```<?php
if (!defined('TYPO3_MODE')) {
	die ('Access denied.');
}
$_EXTKEY = 'xyz';

\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addStaticFile($_EXTKEY, 'Configuration/TypoScript', 'Label');`

Execute this line "vendor/bin/rector process packages/xyz/Configuration/TCA/Overrides/sys_template.php --dry-run -vvv"

Result
```
<?php
if (!defined('TYPO3_MODE')) {
	die ('Access denied.');
}
'Overrides' = 'xyz';

\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addStaticFile('Overrides', 'Configuration/TypoScript', 'Label');
´

I see two things that we should keep in mind when it comes to dynamisation:
1.) Don't replace a simple assign. Maybe we can limit it to functions or/and arrays? I understand that the code here in my example is wrong. So please see it as a basis for discussion ;-)

2.) I don't know if "\basename($fileInfo->getPath());" is a valid solution? Unfortunately, I don't know how flexible the file system is. But in theory we would have to search for ext_emconf.php and use the parent folder from there. Maybe with a recursive function?

Looking forward to feedback :-)

sabbelasichon:
Yes, sure this is possible. Will have a look

[...]

PR #2163, opened 2021-04-10
[FEATURE] Enable EXTKEY to be replaced in sys_template

helsner:
Refs #2162


helsner:
This wouldve failed as i missed out my last code style changes.. wait a sec

PR #2164, opened 2021-04-10
[FEATURE] Enable EXTKEY to be replaced in sys_template

helsner:
Refs #2162

### DIFF
--- a/src/Helper/FileHelperTrait.php
+++ b/src/Helper/FileHelperTrait.php
@@ -18,4 +18,9 @@ trait FileHelperTrait
     {
         return Strings::endsWith($fileInfo->getFilename(), 'ext_tables.php');
     }
+
+    private function isExtEmconf(SmartFileInfo $fileInfo): bool
+    {
+        return Strings::endsWith($fileInfo->getFilename(), 'ext_emconf.php');
+    }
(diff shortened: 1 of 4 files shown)


sabbelasichon:
Thanks. Maybe it can be useful for every file under Configuration/TCA/Overrides. Right? Do you think it is possible to reliable fix that?

helsner:
I was thinking about that, too.
Basically i would guess the replace makes sense in every file, not limited to TCA.
I could think of usages where you have a BE module that requires custom JS which is then used via extPath..
A more basic usage would be for iconfiles in normal TCA config.

So we actually could think about overall dropping the limitation by filename - i dunno if this would cause any trouble though.
We might ask others on this

sabbelasichon:
It is only obligatory be only in ext_emconf.php. At least for extensions in the TER

helsner:
Then i guess we should actually think about dropping the file limitation
What do you think?
We could address this in the channel for more input

sabbelasichon:
Just take the file ext_emconf.php into account and do not substitute it there. On all other files you can safely substitute it. The only thing is if someone use this variable in his custom code. Should rarely be the case, but never say never.

Issue #2177, opened 2021-04-12
Refactor ApplicationFileProcessor


sabbelasichon:
The non php file processor api is completely refactored.

PR #2178, opened 2021-04-12
[TASK] Refactor ApplicationFileProcessors api

sabbelasichon:
Resolves: #2177


src/FlexForms/FlexFormsProcessor.php
```diff
@@ -73,4 +63,33 @@ public function getSupportedFileExtensions(): array
...
+
+    private function processFile(File $file): void
+    {
+        $smartFileInfo = $file->getSmartFileInfo();
+
+        $domDocument = new DOMDocument();
+
+        $domDocument->formatOutput = true;
+
+        $domDocument->load($smartFileInfo->getRealPath());
```
lulco, line 75: Imho you should use File->getContent instead of smartFileInfo->getContent or path because it not working with dry run well

PR #2190, opened 2021-04-18
[WIP] Add failing test for issue #2168

sabbelasichon:


sabbelasichon:
@TomasVotruba I have a weird side effect when i am using my https://github.com/sabbelasichon/typo3-rector/blob/master/src/Rector/Migrations/RenameClassMapAliasRector.php combined with the RenameClassRector. Failing test shows the problem: https://github.com/sabbelasichon/typo3-rector/pull/2190/checks?check_run_id=2375570431. 

If i am changing the line https://github.com/rectorphp/rector/blob/main/src/Configuration/RenamedClassesDataCollector.php#L19 here to 
`php
$this->oldToNewClasses = array_merge($this->oldToNewClasses, ToNewClasses);
`
everything works fine. I cannot really find the issue causing the problem. 

If i am using i.e. the RenameClassRector multiple times in rector itself, everything is fine. 

But the RenameClassAliasMapRector is basically a copy of the RenameClassRector except the configuration part.

Any ideas? What can be the root of the problem.

Issue #2213, opened 2021-04-23
Make icons processor info more precise


helsner:
Currently the `IconsProcessor` says `File "%s" will be moved to %s'`
which results in e.g.
`File "packages/movies" will be moved to packages/movies/Resources/Public/Icons/Extension.gif`

We should adapt this for better understanding

helsner:
I'll do that myself 👌

sabbelasichon:
Thank you.

PR #2218, opened 2021-04-23
[TASK] Correct message of ext icon processing

helsner:
Resolves: #2213


helsner:
I replaced the usage of dirname with the concrete implementation.
This made my rename the vars as it is differed between filepath and directorypath.

I rewrote the rector in my upgrade project to test and overtook it after execution

PR #2228, opened 2021-04-27
[TASK] Support more keywords in globalString and globalVar condition …

Mabahe:
…matcher


sabbelasichon:
@Mabahe Thanks a lot. Could verify the new behaviour with some additional tests here: https://github.com/sabbelasichon/typo3-rector/blob/master/tests/TypoScript/Visitors/OldConditionToExpressionLanguageVisitorTest.php

PR #2327, opened 2021-05-11
Refactor TCA handling to a column-wise approach that enables Tca/Overrides handling

### LINKED ISSUE #2188: Additional report for possible missed out TCA migration addTCAcolumns()
AFAIK we got a new reporting that hints points in the code which cannot fully migrated by rector.

I just found a case in my code which could make use of this:
```
$tmp_showings_columns = [
    'showing' => [
        'exclude' => 1,
        'label' => 'LLL:EXT:offers/Resources/Private/Language/locallang_db.xlf:tx_showings_domain_model_spaceunit.showing',
        'config' => [
            'type' => 'select',
            'foreign_table' => 'tx_showings_domain_model_showing',
            'items' => [
                ['', '']
            ],
            'readOnly' => 1,
        ],
    ],
];
\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addTCAcolumns('tx_offers_domain_model_offer',
    $tmp_showings_columns);
```

The select is not enriched with "selectSingle" as the rector handling this addition ( AddRenderTypeToSelectFieldRector ) checks for full TCA.
As the function does specifically require the columsarray without the wrapping of 'columns' or 'ctrl' which are used to evaluate if the code is actually TCA and should be transformed by rector.

I guess it would be difficult to adapt the rector for this as the conditions are there to prevent transformation of code that is not actually TCA.
So we could filter for the usage of  `addTCAcolumns()` and add that to the reporting

MK-42:
This is a first draft of my thoughts about column-wise tca refactoring.
I chose an easy one to show how the code will look like.

Handling stuff like 'move things into/out of the types section' or other cross-dependencies between other parts of tca and the 'columns' part is still WIP, but please have a look.

I tend to extract some stuff into a base-class such as `AbstractColumnwiseTcaRector`. Or would you prefer to have that in the trait instead @sabbelasichon ?

partially fixes #2188

### DIFF
--- a/src/Rector/Tca/AbstractTcaRector.php
+++ b/src/Rector/Tca/AbstractTcaRector.php
@@ -12,7 +12,7 @@ use Rector\Core\Rector\AbstractRector;
 use Ssch\TYPO3Rector\Helper\TcaHelperTrait;
 
 /**
- * @changelog
+ * Base rector that detects Arrays containing TCA definitions and allows to refactor them
  */
 abstract class AbstractTcaRector extends AbstractRector
 {
--- a/src/Rector/Tca/AbstractTcaRector.php
+++ b/src/Rector/Tca/AbstractTcaRector.php
@@ -13,7 +13,7 @@ use Ssch\TYPO3Rector\Helper\TcaHelperTrait;
 /**
  * @changelog
  */
-abstract class AbstractColumnwiseTcaRector extends AbstractRector
+abstract class AbstractTcaRector extends AbstractRector
 {
     use TcaHelperTrait;
 
@@ -21,6 +21,14 @@ abstract class AbstractColumnwiseTcaRector extends AbstractRector
      * @var string
      */
     protected const TYPE = 'type';
+    /**
+     * @var string
+     */
+    protected const CONFIG = 'config';
+    /**
+     * @var string
+     */
+    protected const LABEL = 'label';
 
     /**
      * @var bool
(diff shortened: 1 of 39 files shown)


src/Rector/v8/v6/AddTypeToColumnConfigRector.php
```diff
@@ -116,4 +77,156 @@ public function getRuleDefinition(): RuleDefinition
...
+    }
+
+    // todo: this should go into a base class
+    private function resolveColumnsDefinition(Node\Expr $columnsDefinition): ?Node
+    {
+        if ($columnsDefinition instanceof Node\Expr\Variable) {
+            // the call to addTcaColumns uses a variable to define the columns to add.
+            // to effectively refactor this argument, we need to find the definition of this variable and refactor that.
+            // as a first-order approximation, we look at the previous statement and hope that the argument is defined there
+            $previousStatement = $columnsDefinition->getAttribute(AttributeKey::PREVIOUS_STATEMENT);
```
MK-42, line 149: the following part will probably need some more sophisticated handling. but i would prefer to see such examples in real life code and adapt from here. what do you think?

MK-42:
@sabbelasichon CI failure seems unrelated.

As we discussed, I chose the Array_ node as the root for refactoring now.
As a next step I will adapt the `AddRenderTypeToSelectFieldRector` to this framing to see if it solves our problem at all.

src/Rector/Tca/AbstractTcaRector.php
```diff
@@ -0,0 +1,169 @@
...
+use PhpParser\Node;
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\ArrayItem;
+use PhpParser\Node\Scalar\String_;
+use Rector\Core\Rector\AbstractRector;
+use Ssch\TYPO3Rector\Helper\TcaHelperTrait;
+
+/**
+ * @changelog
```
sabbelasichon, line 15: You can exclude some classes to having a @changelog. Add it here: https://github.com/sabbelasichon/typo3-rector/blob/main/utils/phpstan/src/Rules/AddChangelogDocBlockForRectorClass.php#L32

tests/Rector/v7/v6/AddRenderTypeToSelectFieldRector/Fixture/add_render_type_in_add_tca_columns.php.inc
```diff
@@ -0,0 +1,46 @@
+<?php
```
sabbelasichon, line 1: Cool. This is really cool. Looking forward to it.

MK-42:
ready to merge

PR #2329, opened 2021-05-12
[WIP] Upgrade to Rector 0.11-next

TomasVotruba:


composer.json
```diff
@@ -12,22 +12,25 @@
...
-        "rector/rector-generator": "^0.1.7",
+        "rector/phpstan-rules": "^0.2.2",
+        "rector/extension-installer": "^0.10.2",
+        "rector/rector-generator": "^0.1.8",
         "symplify/coding-standard": "^9.3",
         "symplify/easy-coding-standard": "^9.3",
         "symplify/phpstan-extensions": "^9.3",
-        "symplify/rule-doc-generator": "^9.3",
         "symplify/phpstan-rules": "^9.3",
-        "rector/rector-phpstan-rules": "^0.1",
```
TomasVotruba, line 29: Just renamed to `"rector/phpstan-rules"`

src/Yaml/Form/Transformer/FormYamlRector.php
```diff
@@ -4,7 +4,7 @@
 
 use Rector\Core\Contract\Rector\RectorInterface;
 
-interface FormYamlTransformer extends RectorInterface
+interface FormYamlRector extends RectorInterface
```
TomasVotruba, line 7: Naming best practise: anythign that inherits from `RectorInterface` should have `Rector` suffix.
It will the orientation in the code with united naming :+1:

tests/Yaml/Form/FormYamlProcessorTest.php
```diff
@@ -2,22 +2,29 @@
 
 namespace Ssch\TYPO3Rector\Tests\Yaml\Form;
 
-use Ssch\TYPO3Rector\Tests\Application\ApplicationFileProcessor\AbstractApplicationFileProcessorTest;
+use Iterator;
+use Rector\Testing\PHPUnit\AbstractRectorTestCase;
+use Symplify\SmartFileSystem\SmartFileInfo;
 
-final class FormYamlProcessorTest extends AbstractApplicationFileProcessorTest
+final class FormYamlProcessorTest extends AbstractRectorTestCase
```
TomasVotruba, line 9: Best pratice for testing: test case should inherit from `AbstractRectorTestCase` and test some change of specific file. This will make sure it's easy to add failing test case and fix it :+1:

src/Yaml/Form/Transformer/FormYamlRector.php
sabbelasichon: Agree. This was historically an in between step before we had the FileProcessors in core.

TomasVotruba:
There is a lot to ugprade and rather cleanup. But we'll make it happen :)

Could I ask you for help with `InitializeExtensionComposerJsonProcessor` - could you separate the `FileProcessor` and `Rector` into 2 classes?

sabbelasichon:
Tell me what to cleanup. I can do that myself. I would really like to know what has to be done in order to use the prefixed version. I know that some parts are still quirky, i.e. the InitializeExtensionComposerJsonProcessor is only an experimental and never used Processor. We can actually drop it.

sabbelasichon:
So if you like just remove the InitializeExtensionComposerJsonProcessor. Really. No need to invest time into it.

tests/Yaml/Form/Fixture/my_form.form.yaml
```diff
@@ -29,3 +29,39 @@ finishers:
...
       attachUploads: true
     identifier: EmailToSender
+-----
+finishers:
+  -
+    options:
+      recipients:
+        foo@domain.com: 'Stadtwerke Bonn'
+        bar@domain.com: Bar
+      subject: 'Kontaktanfrage von Website'
```
TomasVotruba, line 39: Btw, I didn't update these manually. All you need to do is run the test case with UT=1:

```bash
UT=1 vendor/bin/phpunit tests/SomeRectorTest.php
```

And Rector will update the 2nd half of test fixtures for you :)

sabbelasichon: Didn´t know that

tests/Yaml/Form/FormYamlProcessorTest.php
sabbelasichon: Did it for every Processors. This will break yours. Sorry.

sabbelasichon:
You can basically accept my changes if you rebase your branch. I have resolved the AbstractApplicationFileProcessorTestCase. Not necessary anymore. Thanks.

tests/Yaml/Form/FormYamlProcessorTest.php
TomasVotruba: Great!

This PR is now just for probe purposes anyway :) most of work is the processors one.

sabbelasichon:
@TomasVotruba 

Maybe we could add this one to the Rector-Core: https://github.com/sabbelasichon/typo3-rector/blob/main/src/Rector/Composer/ReplacePackageComposerRector.php

TomasVotruba:
@sabbelasichon If it's missing we should add it, yes :+1: 

How does it differ from https://github.com/rectorphp/rector-src/blob/main/rules/Composer/Rector/ReplacePackageAndVersionComposerRector.php ?

TomasVotruba:
Closing as replaced by many smaller PRs

sabbelasichon:
Well no version is required. That´s the only but subtle difference.

TomasVotruba:
I see :+1: . Let's add it under `RenameComposerPackageComposerRector`. Could you send PR?

PR #2334, opened 2021-05-12
Simplify ExtensionComposerRector by using ComposerRectorInterface

TomasVotruba:


src/Rector/Composer/ExtensionComposerRector.php
```diff
@@ -17,7 +17,7 @@
  * @changelog https://docs.typo3.org/m/typo3/reference-coreapi/master/en-us/ExtensionArchitecture/ComposerJson/Index.html#extra
  * @see \Ssch\TYPO3Rector\Tests\Rector\Composer\ExtensionComposerRector\ExtensionComposerRectorTest
  */
-final class ExtensionComposerRector implements ExtensionComposerRectorInterface
+final class ExtensionComposerRector implements ComposerRectorInterface
```
TomasVotruba, line 20: Exactly same behavior is supported in Rector natively with this interface :+1:

tests/Rector/Composer/ExtensionComposerRector/Fixture/update_package_no_extra.json
```diff
@@ -9,15 +9,15 @@
...
+    },
     "extra": {
         "typo3/cms": {
             "extension-key": "_temp_fixture_easy_testing"
         }
-    },
-    "type": "typo3-cms-extension",
-    "require": {
-        "typo3/cms-core": "^9.5"
     }
```
TomasVotruba, line 22: Rector ComposerJson printer is using more standardized order of keys. E.g `require` is in the top.

src/Rector/Composer/ExtensionComposerRector.php
sabbelasichon: Yes and no. I wanted to separate these to not have any side effect. Please don´t change it.

sabbelasichon: Wait. I am gonna think about it twice.

TomasVotruba: It's 95 % of duplicated code from Rector internals with some prefixed classes. This API is not public and is already covered by Rector.

What exact side effects? We should fix those if there are some troubles

sabbelasichon: I guess you are right. There were a reason historically. But i think it should be fine now as we have FileProcessors in place. Can´t  remember exactly. So seems good for now.

sabbelasichon:
The change done by Rector should not happen. Otherwise we have a prefixed namespace in the prefixed version. Fixed this a minute ago.

src/Rector/Composer/ExtensionComposerRector.php
TomasVotruba: If there are any bugs, let me know. It's definitely wrong in Rector :)

TomasVotruba:
> The change done by Rector should not happen. Otherwise we have a prefixed namespace in the prefixed version. Fixed this a minute ago.

Will fix :+1:

src/Rector/Composer/ExtensionComposerRector.php
sabbelasichon: I think it should be fine now. There was a time with the composer.json be part of ProcessCommand deep inside. This is gone now.

TomasVotruba:
Ready to merge :heavy_check_mark: :partying_face:

PR #2335, opened 2021-05-12
[phpstan] bunch of static fixes

TomasVotruba:
PHPStan can help us a lot with maintaining code and checking types.
It also discovers bugs with static reflection, so I put some time to get rid of few ignored messages :+1:

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -34,7 +34,6 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             __DIR__ . '/../src/ValueObject',
             __DIR__ . '/../src/TypoScript/Conditions',
             __DIR__ . '/../src/TypoScript/Visitors',
-            __DIR__ . '/../src/Yaml/Form/Transformer',
             __DIR__ . '/../src/FlexForms/Rector',
             __DIR__ . '/../src/Resources/Icons/IconsProcessor.php',
         ]);
--- a/config/config.php
+++ b/config/config.php
@@ -35,6 +35,7 @@ return static function (ContainerConfigurator $containerConfigurator): void {
             __DIR__ . '/../src/TypoScript/Conditions',
             __DIR__ . '/../src/TypoScript/Visitors',
             __DIR__ . '/../src/FlexForms/Rector',
+            __DIR__ . '/../src/Yaml/Form/Rector',
             __DIR__ . '/../src/Resources/Icons/IconsProcessor.php',
         ]);
 
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -77,16 +77,22 @@ parameters:
             message: '#Namespace "Ssch\\TYPO3Rector\\Yaml\\Form\\Rector" is only reserved for "Rector"\. Move the class somewhere else#'
             path:  src/Yaml/Form/Rector/EmailFinisherRectorInterface
 <<<<<<< HEAD
+<<<<<<< HEAD
 =======
+=======
+>>>>>>> 51918729 (static fixes)
 
         # probably bug in Rector generics annotation
         - '#Method Ssch\\TYPO3Rector\\Rector\\v9\\v0\\IgnoreValidationAnnotationRector\:\:refactorValidation\(\) has parameter \$phpDocInfo with generic class Rector\\BetterPhpDocParser\\PhpDocInfo\\PhpDocInfo but does not specify its types\: TNode#'
 
         # node finder
         - '#Method Ssch\\TYPO3Rector\\Rector\\v11\\v0\\ForwardResponseInsteadOfForwardMethodRector\:\:extractForwardMethodCalls\(\) should return array<PhpParser\\Node\\Expr\\MethodCall\> but returns array<PhpParser\\Node\>#'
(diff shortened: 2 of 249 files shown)


config/config.php
```diff
@@ -34,7 +34,6 @@
             __DIR__ . '/../src/ValueObject',
             __DIR__ . '/../src/TypoScript/Conditions',
             __DIR__ . '/../src/TypoScript/Visitors',
-            __DIR__ . '/../src/Yaml/Form/Transformer',
```
TomasVotruba, line 37: This directory does not exist anymore

sabbelasichon: Yes, should be Rector instead of Transformer. PhpStorm is not refactoring this. Shit. Also i found a naming issue with EmailFinisherRectorInterface

src/FlexForms/Rector/RenderTypeFlexFormRector.php
```diff
@@ -18,7 +18,8 @@ final class RenderTypeFlexFormRector implements FlexFormRectorInterface
     public function transform(DOMDocument $domDocument): bool
     {
         $xpath = new DOMXPath($domDocument);
-        /** @var DOMNodeList|DOMElement[] $elements */
+
+        /** @var DOMNodeList<DOMElement> $elements */
```
TomasVotruba, line 22: This is PHPStan syntax for what was here before. 

Basically "`DOMNodeList` is array collection object of `DOMElement` object"

config/config.php
sabbelasichon: Should be EmailFinisherRector without the Suffix. PhpStorm was not accurate here.

src/Rector/v8/v6/RefactorTCARector.php
```diff
@@ -167,21 +167,25 @@ private function addFieldControlInsteadOfWizardsAddListEdit(Array_ $items): void
                 continue;
             }
 
-            foreach ($fieldValue->value->items as $configValue) {
-                if (null === $configValue) {
+            $fieldValueArray = $fieldValue->value;
+            foreach ($fieldValueArray->items as $configValue) {
+                if (! $configValue instanceof ArrayItem) {
```
TomasVotruba, line 172: If we have `SomeType|null`, the exact match to `! not SomeType` will make IDE and phpstan smarted. If this condition passes, it can be only `SomeType`

vs `$value !== null` means only that value is anything except null - weaker

src/Rector/v8/v6/RefactorTCARector.php
```diff
@@ -167,21 +167,25 @@ private function addFieldControlInsteadOfWizardsAddListEdit(Array_ $items): void
...
                     continue;
                 }
 
                 if (! $configValue->value instanceof Array_) {
                     continue;
                 }
 
+                /** @var Array_ $configValueArray */
+                $configValueArray = $configValue->value;
```
TomasVotruba, line 181: This is sometimes useful, because PHPStan/IDE can lose type of single nested variables.

src/Rector/v9/v0/InjectAnnotationRector.php
```diff
@@ -73,11 +72,7 @@ public function refactor(Node $node): ?Node
         $injectMethods = [];
         $properties = $node->getProperties();
         foreach ($properties as $property) {
-            /** @var PhpDocInfo|null $propertyPhpDocInfo */
             $propertyPhpDocInfo = $this->phpDocInfoFactory->createFromNodeOrEmpty($property);
```
TomasVotruba, line 75: This alwasy returns an `PhpDocInfo` object, even for empty doc

sabbelasichon: Yeah. You do smarter analysis than me. This was some old code.

sabbelasichon:
@TomasVotruba Be aware that the FileFormatter for xml does some modification. Maybe they are not super nice, but it is better than the default printing from DomDocument.

config/config.php
TomasVotruba: Yea, luckily we have PHPStan and Rector to cover us :)

TomasVotruba:
@sabbelasichon Should be ok now :+1: 

Ready to review

config/config.php
sabbelasichon: Fixed it in a follow up commit.

phpstan.neon
```diff
@@ -21,15 +21,8 @@ parameters:
          - tests/bootstrap.php
 
     inferPrivatePropertyTypeFromConstructor: true
-    checkMissingIterableValueType: false
```
sabbelasichon, line 24: Yeah. Good catch.

sabbelasichon:
I would say Ready to merge if you think so too.

config/config.php
TomasVotruba: I'll add the correct ignore and suffix in next PR. This is already big one

sabbelasichon: I have done that already.

TomasVotruba: Ok, I'll rebase :)

PR #2336, opened 2021-05-12
[phpstan] more fixes

TomasVotruba:


sabbelasichon:
I have added also a label "Ready to merge".

src/TypoScript/Visitors/ExtbasePersistenceVisitor.php
```diff
@@ -149,7 +150,9 @@ private function extractSubClasses(array $paths, Assignment $statement): void
             ];
         }
 
-        self::$persistenceArray[$className][self::SUBCLASSES][] = $statement->value->value;
+        /** @var ScalarValue $scalarValue */
+        $scalarValue = $statement->value;
+        self::$persistenceArray[$className][self::SUBCLASSES][] = $scalarValue->value;
```
TomasVotruba, line 155: This is a tricky one. The class in typoescript traverser is called `Scalar`.
But `Scalar` will be soon PHP native type. So `scalar` is now percieved as `int|string|bool|float` type. 

To make a difference, I've used `ScalarValue` types. IMO this package should rename the class as soon possible to avoid reserved type mayhem (it's hell to upgrade reserved types on version where they're already reserved :

sabbelasichon: https://github.com/martin-helmich/typo3-typoscript-parser/pull/64

PR #2344, opened 2021-05-13
Move special configuration to columns overrides

### LINKED ISSUE #1988: Migrate special configuration and remove show item style pointer config
https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/8.6/Deprecation-79440-TcaChanges.html

Basically we have to write a rector for this method https://github.com/TYPO3/TYPO3.CMS/blob/9e1c91aed86490f75730ae186504bd900057107e/typo3/sysext/core/Classes/Migrations/TcaMigration.php#L184

MK-42:
fixes #1988 

The tests for the core migration take into account that another migration is run after the one handled here.
To separate responsibilities and keep the code close to the core migration, i chose to adapt the fixtures and add some test-cases to the second rector that ensures that we finally reach the desired state after running both of them.


sabbelasichon:
Really really good stuff. Keep it up. It is so readable now.

PR #2354, opened 2021-05-15
[WIP] config: remove local phpstan config reference

TomasVotruba:
It would cause trouble like these 
https://github.com/rectorphp/rector-src/pull/46/checks?check_run_id=2592322406#step:5:15

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -46,6 +46,9 @@
         },
         "exclude-from-classmap": [
             "**.php.inc"
+        ],
+        "classmap": [
+            "stubs"
         ]
     },
     "config": {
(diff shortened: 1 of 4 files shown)


sabbelasichon:
The stubs should be added to the composer.json classmap like here: https://github.com/rectorphp/rector-src/blob/main/composer.json#L88

TomasVotruba:
That's the best. Yes :+1: 
Just in `autoload-dev`, so it's local for your project

sabbelasichon:
Locally everything went fine. Why are the stubs not found and static reflection is not working in the ci?

sabbelasichon:
Ok. I am gonna fix it. I have seen what you have done for nette.

PR #2358, opened 2021-05-16
Allow refactoring TCA section 'ctrl' using AbstractTcaRector and rework types-refactoring a bit

MK-42:


MK-42:
@sabbelasichon CI is not triggered anymore - is that on purpose, do do I need to adapt something here?

MK-42:
ready for review.
I'm skimming through some of the tca rectors to find blindspots that are not yet covered with the new structure. I would then write some .rst about the whole thing - i think that might help other devs to chime in here.

are there any other not-yet-implemented tca migrations that i should tackle along the way?

sabbelasichon:
Thanks. No, not i am aware of. I have labeled all of them with TCA or TCA Core. We should be done for now.

sabbelasichon:
@MK-42 Thanks a lot. Looks clean. It is getting brighter and brighter

Issue #2420, opened 2021-05-26
Method $GLOBALS['TT']->start(); is wrongly transformed


sabbelasichon:
if (!is_object($GLOBALS['TT'])) {
            $GLOBALS['TT'] = new \TYPO3\CMS\Core\TimeTracker\NullTimeTracker;
            $GLOBALS['TT']->start();
        }

PR #2421, opened 2021-05-28
[TASK] Add failing test

sabbelasichon:
Relates: #2420


sabbelasichon:
If we use a instantiate a class without parentheses and try to add arguments to the constructor the output is really weird / is broken.

Issue #2410, opened 2021-05-25
RefactorProcessOutputRector.php Missing Response Parameter. 


tomasnorre:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.10.6 (PHP 7.4)             |

To me, it looks like that the `RefactorProcessOutputRector` should a `Response` as input. 

https://github.com/sabbelasichon/typo3-rector/blob/main/src/Rector/v9/v5/RefactorProcessOutputRector.php#L89

https://github.com/TYPO3/TYPO3.CMS/blob/master/typo3/sysext/frontend/Classes/Controller/TypoScriptFrontendController.php#L2816

https://review.typo3.org/c/Packages/TYPO3.CMS/+/58499/10/typo3/sysext/frontend/Classes/Http/RequestHandler.php

## Minimal PHP Code Causing Issue

```php
$tsfe = GeneralUtility::makeInstance(TypoScriptFrontendController::class);
$tsfe->processOutput();
``` 

Will result in: 

```php
$tsfe = GeneralUtility::makeInstance(TypoScriptFrontendController::class);
$tsfe->applyHttpHeadersToResponse();
$tsfe->processContentForOutput();
```
## Expected Behaviour

I would have expected it to be: 

```php
$tsfe = GeneralUtility::makeInstance(TypoScriptFrontendController::class);
$tsfe->applyHttpHeadersToResponse(new Response());
$tsfe->processContentForOutput();
```

PR #2422, opened 2021-05-28
[BUGFIX] Add Response object

sabbelasichon:
Resolves: #2410


sabbelasichon:
@tomasnorre Is this enough? The method seems only used once in the core during the bootstrapping of the frontend.

tomasnorre:
From my experience yes, it's enough.

Issue #2418, opened 2021-05-26
Bug: initializeArguments duplication for ViewHelper overrides


helsner:
# Bug Report
Rector Version 0.10.x prefixed or dev-main 

ViewHelpers that are overrides of core viewhelpers break due to the `MoveRenderArgumentsToInitializeArgumentsMethodRector` adding the `initializeArguments` as second time based on the render method.
Exception
`Argument "<ARGUMENT>" has already been defined, thus it should not be defined again.`

## Minimal PHP Code Causing Issue

```
<?php

declare(strict_types=1);

namespace DFAU\Test\Overrides;

use TYPO3\CMS\Fluid\ViewHelpers\Link\PageViewHelper;

class LinkPageViewHelperOverride extends PageViewHelper
{
    /**
     * @param int|null $pageUid target page. See TypoLink destination
     * @param array $additionalParams query parameters to be attached to the resulting URI
     * @param int $pageType type of the target page. See typolink.parameter
     * @param bool $noCache set this to disable caching for the target page. You should not need this.
     * @param bool $noCacheHash set this to suppress the cHash query parameter created by TypoLink. You should not need this.
     * @param string $section the anchor to be added to the URI
     * @param bool $linkAccessRestrictedPages if set, links pointing to access restricted pages will still link to the page even though the page cannot be accessed
     * @param bool $absolute If set, the URI of the rendered link is absolute
     * @param bool $addQueryString If set, the current query parameters will be kept in the URI
     * @param array $argumentsToBeExcludedFromQueryString arguments to be removed from the URI. Only active if $addQueryString = TRUE
     * @param string $addQueryStringMethod Set which parameters will be kept. Only active if $addQueryString = TRUE
     *
     * @return string Rendered page URI
     */
    public function render($pageUid = null, array $additionalParams = [], $pageType = 0, $noCache = false, $noCacheHash = false, $section = '', $linkAccessRestrictedPages = false, $absolute = false, $addQueryString = false, array $argumentsToBeExcludedFromQueryString = [], $addQueryStringMethod = null)
    {
        // just a test
        return;
    }
}

```

## Expected Behaviour

If there is a chance to fix this then i guess via override detection.
We would have to detect if the parent class is an abstract class or not.
If not then the rector should report and skip.
We cannot throw away the arguments as there might be real custom arguments that got added via override.
At the same time we cannot keep the duplicated ones which seems hard to resolve in general.
[...]

PR #2423, opened 2021-05-28
[TASK] Add failing test

sabbelasichon:
Relates: #2418


sabbelasichon:
@helsner Tough one. But Rector itself has now an AstResolver, so we can use the Reflection and the PhpParse in combination to reflect and analyze parent methods, classes etc.

Issue #2432, opened 2021-06-02
ApplicationContextConditionMatcher rewrite is invalid


helsner:
# Bug Report

latest version

## Minimal PHP Code Causing Issue

`applicationContext=Development`
is rewritten into
`applicationContext==Development`

which is actually invalid as the quotes are missing
`applicationContext=="Development"`
would be valid

helsner:
i try to patch that

helsner:
somehow my test won't start locally, so i push to see if its my issue..

PR #2433, opened 2021-06-02
[BUGFIX] Add missing quotes to ApplicationContextConditionMatcher

helsner:
Resolves: #2432


helsner:
somehow my test won't start locally, so i push to see if its my issue..

Issue #2462, opened 2021-06-19
Refactor logicalOr/logicalAnd with more than one constraint into an array


ayacoo:
# Feature Request

Context: Upgrade from TYPO3 version 8 to version 9

In older projects we have logicalOr and logicalAnd with n constraints. In newer TYPO3 versions these functions allowed only one array with these contraints or just one constraint: "mixed"

function description

    /**
     * Performs a logical disjunction of the two given constraints. The method
     * takes one or more constraints and concatenates them with a boolean OR.
     * It also accepts a single array of constraints to be concatenated.
     *
     * @param mixed $constraint1 The first of multiple constraints or an array of constraints.
     * @return OrInterface
     */
    public function logicalOr($constraint1);

Source file: public/typo3/sysext/extbase/Classes/Persistence/QueryInterface.php

## Current code

```
        $constraints[] = $query->logicalOr(
            $query->equals('uid', $product->getUid()),
            $query->equals('parentProduct', $parentProduct),
            $query->equals('uid', $parentProduct->getUid())
        );
```

## Expected code

```
        $constraints[] = $query->logicalOr([
            $query->equals('uid', $product->getUid()),
            $query->equals('parentProduct', $parentProduct),
            $query->equals('uid', $parentProduct->getUid())
        ]);
```

## Possibile logic
If logicalOr/logicalAnd have more than 1 argument, we could move them to an array

## Additional info
TYPO3 Rector currently does nothing with the arguments, but Rector uses this class: "RemoveExtraParametersRector" I don't know if we can prioritise higher with this sabbelasichon/typo3-rector than with the original rector/rector package. 

Unfortunately, I haven't found a changelog entry where we can classify this. Maybe someone knows more? In case of doubt, this problem is simply documented now ;-)

sabbelasichon:
@ayacoo It is not really possible to define the order here. But i will try to investigate more to see if this is reliable fixable.

ayacoo:
@sabbelasichon Thank you for your feedback. Please don't invest too much time here. It is a rather old construct, which is certainly hardly used any more. If it is not easy, please close it directly.

PR #2477, opened 2021-06-24
[TASK] Use array for logicalAnd and logicalOr query calls

sabbelasichon:
Resolves: #2462


sabbelasichon:
@ayacoo Seems to work nicely if you put the RemoveExtraParametersRector at the end of the rules.

sabbelasichon:
But to be honest, when using Rector i would use the destructive rules after everything else.

Issue #2456, opened 2021-06-14
Catch TypoScript parsing errors && log them


helsner:
Currently we got some false positive issues with e.g. typescript files in build folders etc.
To prevent typoscript parsing issues as those files are wrongfully taken into account we want to log the parsing issue instead of throwing the exception.
Also the default template should be extended to skip directories like node_modules or similar in the Resources directory to prevent the overall parsing of files that are not typoscript

PR #2482, opened 2021-06-28
[TASK] Change ts parse error message to warning, add file path

helsner:
Resolves: #2456


sabbelasichon:
@helsner I am gonna fix the PHPStan error here.

PR #2486, opened 2021-06-30
Fix ClassInsertManipulator::addAsFirstMethod() cannot pass Nop instance

samsonasik:
@sabbelasichon I think this can be solution for error https://github.com/rectorphp/rector-src/pull/338/checks?check_run_id=2949192987#step:5:24 

ref https://github.com/rectorphp/rector-src/pull/338#pullrequestreview-695721105

what do you think?


sabbelasichon:
@samsonasik It looks good. Let´s see what the tests are saying.

sabbelasichon:
@samsonasik Thank you for the quick fix. I am gonna create a new release now.

PR #2497, opened 2021-07-04
use stable rector

TomasVotruba:


TomasVotruba:
Green is back :heavy_check_mark:

Issue #2491, opened 2021-07-01
$GLOBALS['TSFE']->renderCharset getting replaced in assignment


helsner:
latest version

$GLOBALS['TSFE']->renderCharset = 'utf-8';

## Minimal PHP Code Causing Issue
```
- $GLOBALS['TSFE']->renderCharset = 'utf-8';
+'utf-8' = 'utf-8';

```
## Expected Behaviour
Nothing to happen

helsner:
I'll try that one

helsner:
ok i'm kinda lost on this one :D i don't get anything on how to solve
i would extend the `shouldSkip` check, but i don't know how to check for this

sabbelasichon:
@helsner No problem. I will create a Pull Request and assign you as a reviewer.

PR #2500, opened 2021-07-07
[BUGFIX] Do not change assignment for renderCharset

sabbelasichon:
Resolves: #2491


sabbelasichon:
@helsner That´s how i solved the issue here.

PR #2504, opened 2021-07-08
Make extensive use of PHP8/7.4

joernneumeyer-beech:
This pull request aims to replace old code snippets with new ones.
Doing so is desirable, because it leads to more concise and semantically better speaking code.

The following upgrades have taken place (where applicable):
- replace `strpos` with `str_contains`, in cases where it has been used to check if a string contains another string
- replace `switch` and `if/else`-chains with `match`-expressions, in cases where they have been used to return different values based on some input value
- replace anonymous functions with arrow functions, in cases where they immediately return a result and don't do further processing/checks

### DIFF
--- a/src/Helper/TcaHelperTrait.php
+++ b/src/Helper/TcaHelperTrait.php
@@ -70,12 +70,7 @@ trait TcaHelperTrait
 
     protected function extractArrayValueByKey(?Node $node, string $key): ?Expr
     {
-        $item = $this->extractArrayItemByKey($node, $key);
-        if (null === $item || null === $item->value) {
-            return null;
-        }
-
-        return $item->value;
+        return $this->extractArrayItemByKey($node, $key)?->value;
     }
 
     protected function hasKeyValuePair(Array_ $configValue, string $configKey, string $expectedValue): bool
@@ -192,7 +187,7 @@ trait TcaHelperTrait
 
             // search the config sub-array for this field
             foreach ($columnConfig->value->items as $configValue) {
-                if (null === $configValue || null === $configValue->key) {
+                if (null === $configValue?->key) {
                     continue;
                 }
 
--- a/src/FileProcessor/Composer/Rector/ExtensionComposerRector.php
+++ b/src/FileProcessor/Composer/Rector/ExtensionComposerRector.php
@@ -147,7 +147,7 @@ CODE_SAMPLE
             return;
         }
 
-        if (false === strpos($name, '_')) {
+        if (!str_contains($name, '_')) {
             return;
         }
 
--- a/src/FileProcessor/TypoScript/Conditions/GlobalStringConditionMatcher.php
+++ b/src/FileProcessor/TypoScript/Conditions/GlobalStringConditionMatcher.php
@@ -45,7 +45,8 @@ final class GlobalStringConditionMatcher extends AbstractGlobalConditionMatcher
                 'IENV' => $this->createIndependentCondition($property, $operator, $value),
                 'TSFE' => $this->refactorTsfe($property, $operator, $value),
                 'GP' => $this->refactorGetPost($property, $operator, $value),
-                'LIT' => sprintf('"%s" %s "%s"', $value, self::OPERATOR_MAPPING[$operator], $property)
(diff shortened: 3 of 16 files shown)


sabbelasichon:
@joernneumeyer-beech Thanks. Just a note why it was not done before: The rector/rector is a downgraded version and the str_contains was not part of the Downgrade sets before. I think it is now, but we still have to be using the new shiny features wisely. If you have checked it before, awesome. If you didn´t know the process of downgrading the *-src i hopefully could add some light.

sabbelasichon:
Btw. did you use rector for that?

sabbelasichon:
Could you fix the phpstan errors? The coding standards workflow is broken atm. So, you don´t have to care about this one.

joernneumeyer-beech:
Thanks for the info about the downgrading, I'll keep it in mind.

No, I did the refactor by hand. I could try adding a rector for the things I did though.

I'll have a look at the phpstan errors.

docs/configuration_and_processing.md
```diff
@@ -34,11 +34,9 @@ use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigura
 return static function (ContainerConfigurator $containerConfigurator): void {
     $parameters = $containerConfigurator->parameters();
 
-    $containerConfigurator->import(Typo3SetList::TYPO3_76);
-    $containerConfigurator->import(Typo3SetList::TYPO3_87);
-    $containerConfigurator->import(Typo3SetList::TYPO3_95);
-    $containerConfigurator->import(Typo3SetList::TYPO3_104);
-    $containerConfigurator->import(Typo3SetList::TYPO3_11);
+    $containerConfigurator->import(Typo3SetList::TCA_76);
```
sabbelasichon, line 37: Why is this change done?

src/Helper/TcaHelperTrait.php
```diff
@@ -192,7 +187,7 @@ private function extractColumnConfig(Array_ $items, string $keyName = 'config'):
 
             // search the config sub-array for this field
             foreach ($columnConfig->value->items as $configValue) {
-                if (null === $configValue || null === $configValue->key) {
+                if (null === $configValue?->key) {
```
sabbelasichon, line 190: I am not sure if this is supported by the downgrade sets. I have to check this.

sabbelasichon:
@joernneumeyer-beech Try to run the rector rules for your changes. 

https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#closuretoarrowfunctionrector
https://github.com/rectorphp/rector/blob/main/docs/rector_rules_overview.md#changeswitchtomatchrector

sabbelasichon:
But as i said, we are doing an automatic downgrade with rector itself, we should still be careful in some areas. I have to check this. Even if the changes seem simple it is not, because they could break the downgrade process. Let me check. 
But i am on holiday till 1st of August. So please bear with me.

docs/configuration_and_processing.md
joernneumeyer-beech: My bad. I first did a pull from `master`. I didn't see that the default branch is `main`. I'll fix that.

joernneumeyer-beech:
I ran Rector with the two rules added and it only applied the rule `NewlineBeforeNewAssignSetRector` for the file `src/FileProcessor/TypoScript/Rector/OldConditionToExpressionLanguageTypoScriptRector.php` between the two `array_map` calls (for language level 7.2).

sabbelasichon:
@joernneumeyer-beech What will happen if you raise the PHP-Version in your configuration to level 8.0. That should apply the mentioned rules. You could even try to use the whole sets for PHP 7.4 and PHP 8. That´s what i should do in my configuration anyways ;-)

https://github.com/rectorphp/rector-src/blob/main/rector.php#L37

sabbelasichon:
@joernneumeyer-beech Thanks. We are going to merge it.

src/Helper/TcaHelperTrait.php
sabbelasichon: This is now supported with DowngradeNullsafeToTernaryOperatorRector

TomasVotruba:
@joernneumeyer-beech Very nice work :clap:

PR #2506, opened 2021-07-09
Bump deps to nikic/php-parser 4.11

TomasVotruba:
News https://github.com/nikic/PHP-Parser/releases/tag/v4.11.0

Attempt to fix https://github.com/rectorphp/rector-src/pull/410#issuecomment-877346778


TomasVotruba:
Merging as CS is stuck for 8 minutes and is passing locally. We need to fix rector-src

PR #2562, opened 2021-09-10
[TASK] Add ExtbasePersistenceTypoScriptRector to v10 set

helsner:


helsner:
I'm not sure about adding a config on the rector inclusion?
Also on changing the readme + default rector file.
The declaration there would be obsolete as we have it in the base set then

PR #2654, opened 2021-10-02
[phpstan] fix phpstan errors caused by new php-parser and symplify releases

TomasVotruba:
I've noticed the CI is failing here because of php-parser and mine changes, let's make it green again :)


sabbelasichon:
Thank you very much

Issue #2600, opened 2021-09-24
Deprecation: #94309 - GeneralUtility::stdAuthCode


sabbelasichon:
Deprecation: #94309 - GeneralUtility::stdAuthCode

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Deprecation-94309-DeprecatedGeneralUtilitystdAuthCode.html
.. include:: ../../Includes.txt

=================================================
Deprecation: #94309 - GeneralUtility::stdAuthCode
=================================================

See :issue:`94309`

Description
===========

The method :php:`\TYPO3\CMS\Core\Utility\GeneralUtility::stdAuthCode()`
has not been used within the Core since at least v9. It internally fiddles
with the `encryptionKey` while using :php:`md5()`. Furthermore, the default
length of 8 chars could easily lead to hash collisions. The TYPO3 Core already
provides :php:`\TYPO3\CMS\Core\Utility\GeneralUtility::hmac()` for such
purposes, which is using `sha1` with a length of 40. Therefore,
:php:`stdAuthCode()` has been deprecated and will be removed in TYPO3 v12.

Impact
======

Calling the method will trigger a PHP :php:`E_USER_DEPRECATED` error.

Affected Installations
======================

All TYPO3 installations calling this method in custom code. The extension
scanner will find all usages as strong match.

Migration
=========

Replace all usages of the method in custom extension code by either using
:php:`\TYPO3\CMS\Core\Utility\GeneralUtility::hmac()` or by a custom
implementation.

.. index:: PHP-API, FullyScanned, ext:core

helsner:
@sabbelasichon i guess i need some help, not because of rector but because of the actual migration.
stdAuthCode can be used with integer or even arrays.
hmac only works with strings.
That means we have to transform the array into a string, but i wonder in which format that should end up?

Also stdAuthCode supports the addition of a length argument, which hmac does not.
For this i guess we could use the `$authCode = substr(md5($authCode), 0, $codeLength);` of the original stdAuthCode - but i want to mention here, so its documented and for myself 😄

helsner:
ok, it just came to me, that i can use everything of the old method for handling the fields etc.
my bad - i got it

helsner:
@sabbelasichon guess i need help on that one.
I created the test cases that we need to transform, but from my point of view we need a function to handle the difference in types

sabbelasichon:
[...]

PR #2655, opened 2021-10-02
[FEAT] Replace stdAuthCode with Hmac rector

helsner:
Resolves: #2600


helsner:
@sabbelasichon i dont quite understand what to do with this 😬

tests/Rector/v11/v3/ReplaceStdAuthCodeWithHmacRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,15 @@
+<?php
+
+namespace Ssch\TYPO3Rector\Tests\Rector\v11\v3\ReplaceStdAuthCodeWithHmacRector\Fixture;
+
+\TYPO3\CMS\Core\Utility\GeneralUtility::stdAuthCode(5);
+
+?>
+-----
```
sabbelasichon, line 8: If it is not touched remove the code under the dashes.

sabbelasichon (changes requested):

tests/Rector/v11/v3/ReplaceStdAuthCodeWithHmacRector/Fixture/fixture.php.inc
helsner: changed

helsner:
@sabbelasichon should be fine this way - those tests seem not to be broken by my change but in general?

PR #2657, opened 2021-10-04
bump phpstan dev rules

TomasVotruba:
Hey, here is an update for changes in loading of PHPStan configs :partying_face: 

We're moving to use phpstan/extension-installer if present, for PHPStan that runs under Rector. That way people can use extension installer and do not need to duplicate include configs in `phpstan-for-rector.neon` configs

Contributed in `rector/rector-src` here
https://github.com/rectorphp/rector-src/pull/952/commits/7ef28cab4333f3d914bfeacf96660ecb0089904e


TomasVotruba:
Mergin to make it :heavy_check_mark:  and test in `rector/rector-src` :+1:

Issue #2601, opened 2021-09-24
Deprecation: #94311 - GeneralUtility::rmFromList


sabbelasichon:
Deprecation: #94311 - GeneralUtility::rmFromList

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Deprecation-94311-DeprecatedGeneralUtilityrmFromList.html
.. include:: ../../Includes.txt

================================================
Deprecation: #94311 - GeneralUtility::rmFromList
================================================

See :issue:`94311`

Description
===========

The method :php:`\TYPO3\CMS\Core\Utility\GeneralUtility::rmFromList()` has not
been used in the Core since v10. The method has now been deprecated
in :php:`\TYPO3\CMS\Core\Utility\GeneralUtility` and will be removed in
TYPO3 v12.

Impact
======

Calling the method will trigger a PHP :php:`E_USER_DEPRECATED` error.

Affected Installations
======================

All TYPO3 installations calling this method in custom code. The extension
scanner will find all such usages as strong match.

Migration
=========

Replace all usages of the method in your extension code.

.. index:: PHP-API, FullyScanned, ext:core

sabbelasichon:
implode(',', array_filter(explode(',', $list), function($item) use($element) {
           return $element == $item;
        }));

helsner:
lol i was on the same one right now.. :D thanks though! yours looks better and probably works 😂

helsner:
@sabbelasichon i need help with the function init via rector

PR #2658, opened 2021-10-05
[TASK] Substitute GeneralUtility::rmFromList

sabbelasichon:
Resolves: #2601


sabbelasichon:
@helsner Here is the solution for #2601

PR #2695, opened 2021-10-07
update resolve() and getStaticType() to united getType() method

TomasVotruba:
Hi Sebastian,

I hope you're great :)

This is just compatibility update for change in Rector: https://github.com/rectorphp/rector-src/pull/973


sabbelasichon:
Thank you for taking care

Issue #2623, opened 2021-09-24
Deprecation: #94956 - Public $cObj


sabbelasichon:
Deprecation: #94956 - Public $cObj

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.4/Deprecation-94956-PublicCObj.html
.. include:: ../../Includes.txt

==================================
Deprecation: #94956 - Public $cObj
==================================

See :issue:`94956`

Description
===========

Frontend plugins receive an instance of :php:`\TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer` when
called via :php:`ContentObjectRenderer->callUserFunction()`. This is
typically the case for plugins called as :typoscript:`USER` or indirectly
as :typoscript:`USER_INT` type.

The instance of :php:`ContentObjectRenderer` has previously been set by
declaring a public (!) property :php:`cObj` in the consuming class.

Handing a :php:`ContentObjectRenderer` instance around this way is hard to
follow and has thus been deprecated: Declaring :php:`public $cObj` should
be avoided. Frontend plugins that need the current :php:`ContentObjectRenderer`
should have a public :php:`setContentObjectRenderer()` method instead.

Impact
======

Declaring :php:`public $cObj` in a class called by
:php:`ContentObjectRenderer->callUserFunction()` triggers a PHP :php:`E_USER_DEPRECATED` error.

Affected Installations
======================

Frontend extension classes that neither extend :php:`TYPO3\CMS\Frontend\Plugin\AbstractPlugin`
("pibase") nor Extbase :php:`TYPO3\CMS\Extbase\Mvc\Controller\ActionController`
and have a public property :php:`cObj` are affected.

Migration
=========

When instantiating the frontend plugin, :php:`ContentObjectRenderer->callUserFunction()`
now checks for a public method :php:`setContentObjectRenderer()` to explicitly set
an instance of the :php:`ContentObjectRenderer`.

Many plugins may not need this instance at all. If the ContentObjectRenderer instance
used within the plugin does not rely on further ContentObjectRenderer state, for instance
if it only calls :php:`stdWrap()` or similar without using state like :typoscript:`LOAD_REGISTER`,
the :php:`cObj` class property should be avoided and an own instance of  ContentObjectRenderer
should be created.

Classes that do rely on current ContentObjectRenderer state should adapt their code.

Before::

    class Foo
    {
        public $cObj;
    }

After::

    class Foo
    {
        protected $cObj;

[...]

PR #2716, opened 2021-10-27
feat: handle public $cObj and add set function

helsner:
Refs #2623


helsner:
@TomasVotruba Hi Tomas, i would like to take the chance to get to know you! 😄 
Maybe you got a hint for me as well as its basic rector

I need to change the visibility of the property from `public` to `protected` do i do this by remove and readd or is there another way?

src/Rector/v11/v4/ProvideCObjViaMethodRector.php
```diff
@@ -0,0 +1,116 @@
...
+        if ($cObjProperty->isPublic()) {
+            // TODO: add handling to make it protected
+        }
+
+        $classMethod = $node->getMethod('setContentObjectRenderer');
+        if ($classMethod instanceof ClassMethod) {
+            return null;
+        }
+
+        $this->addSetContentObjectRendererMethod($node);
```
TomasVotruba, line 65: Here the modified class node should be returned, as you're adding an item to array.
That's my rough guess :)

src/Rector/v11/v4/ProvideCObjViaMethodRector.php
```diff
@@ -0,0 +1,116 @@
...
+    /**
+     * @param Class_ $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if ($this->isObjectType($node, new ObjectType('TYPO3\CMS\Frontend\Plugin\AbstractPlugin'))
+            || $this->isObjectType($node, new ObjectType('TYPO3\CMS\Extbase\Mvc\Controller\ActionController'))
+        ) {
+            return null;
+        }
```
TomasVotruba, line 49: So if the class is type A or B, it should be skipped?
Should be inverted? :)

sabbelasichon:
@helsner Thanks a lot, well done. Let´s ship it to the public.

PR #2720, opened 2021-11-02
Require Symplify 10

TomasVotruba:


sabbelasichon:
@TomasVotruba I don´t like the var-exporter construct either. But it was the only way to generate a nicer formatted array for some TYPO3 internal transformation. If i would use the AST-Parser the array would end up a in single line.

TomasVotruba:
I see. Any printer of new value would have such issue, that's why there must be coding standard to handle it :) 

Thing is, I'm trying to cut down unused Symfony dependencies, so we can ship faster and lighter and car about less bugs. I've managed to get rid of symfony cache, event-dispatcher, http-foundation, http-kernel, error-handler and logger.

This is the only call of var-exporter in whole Rector project:
https://github.com/rectorphp/rector/tree/main/vendor/symfony/var-exporter

TomasVotruba:
Main focus is now to upgrade to Symplify 10. The var-exporter can be handled later. It would be nice if you're able to manage without it.

Issue #2721, opened 2021-11-03
Select $_EXTKEY replacement value from composer.json


alexanderschnitzler:
# Feature Request

When replacing `$_EXTKEY` variables, the directory name is used to determine the target value. This however is problematic in setups where local extensions are stored as such:

- Original extension name: `vendor_ext`
- Extension name set in composer.json: `vendor_ext`
- Package name set in composer.json: `vendor/ext`
- Stored in: `packages/vendor/ext`

Now, `$_EXTKEY` is replaced with `ext`, when it should be replaced with `vendor_ext`.

Instead of just evaluating the source package folder name, the value could be taken from:

```json
    "extra": {
        "typo3/cms": {
            "extension-key": "vendor_ext"
        }
    }
```

This even happens when already setting the latter configuration and having correctly named symlinks in `typo3conf/ext/` as the symlink is resolved and again the source folder name is taken into account.

PR #2722, opened 2021-11-03
Make ReplaceExtKeyWithExtensionKeyRector more robust

sabbelasichon:
- Reading from composer.json extension key if exists
- Reading from composer.json name
- last resort use the folder name

Resolves: #2721


src/Rector/v9/v0/ReplaceExtKeyWithExtensionKeyRector.php
```diff
@@ -105,7 +114,29 @@ private function createExtensionKeyFromFolder(SmartFileInfo $fileInfo): ?SmartFi
...
+        try {
+            $composerJson = new SmartFileInfo($extEmConf->getRealPathDirectory() . '/composer.json');
+            $json = json_decode($composerJson->getContents(), true);
+
+            if (isset($json['extra']['typo3/cms']['extension-key'])) {
+                return $json['extra']['typo3/cms']['extension-key'];
+            }
+
+            if (isset($json['name'])) {
+                [, $extensionKey] = explode('/', $json['name'], 2);
```
alexanderschnitzler, line 133: I would implement this differently. Usually the extension key is `vendor_ext`. Any when naming a package, users usually use `vendor/ext`. What you do here is drop the vendor completely and use the name only as possible folder name. This works for popular extensions like `news` and `powermail` where the uniqueness is given by the uniqueness inside the TER but those are usually not the extensions you touch during an update in your instance. It's your own local extensions which are mostly prefixed with a customer or agency prefix aka the vendor. Hence, I would try to combine both parts of the name. vendor and package name.

Issue #2670, opened 2021-10-06
Deprecation: #95219 - TypoScriptFrontendController->ATagParams


sabbelasichon:
Deprecation: #95219 - TypoScriptFrontendController->ATagParams

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.5/Deprecation-95219-TypoScriptFrontendController-ATagParams.html
.. include:: ../../Includes.txt

==============================================================
Deprecation: #95219 - TypoScriptFrontendController->ATagParams
==============================================================

See :issue:`95219`

Description
===========

The public property :php:`TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController->ATagParams`
has been marked as deprecated.

It was used in the past as a copy of the value
:php:`TypoScriptFrontendController->config[config][ATagParams]`,
which should be used instead.

There is no need to use such a (less prominent) configuration option in a
separate public property, as it needs to be kept in sync with the
actual configuration option.

The second argument of the related method
:php:`TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer->getATagParams()`
called :php:`$addGlobal` is also marked as deprecated, and will have no effect
anymore in TYPO3 v12.

Impact
======

Accessing, setting or writing this property will trigger a PHP :php:`E_USER_DEPRECATED` error.

Calling :php:`ContentObjectRenderer->getATagParams()`
with a second argument set to false will trigger a PHP :php:`E_USER_DEPRECATED` error
as well.

Affected Installations
======================

TYPO3 installations with third-party-extensions accessing, or
writing this property directly within PHP, or calling :php:`getATagParams()`
directly, which is highly unlikely.

Migration
=========

All calls of :php:`$GLOBALS['TSFE']->ATagParams` can be replaced
with :php:`$GLOBALS['TSFE']->config['config']['ATagParams'] ?? ''`.

.. index:: Frontend, PHP-API, TypoScript, FullyScanned, ext:frontend

PR #2723, opened 2021-11-03
[FEAT] Replace TSFE->ATagParams calls

helsner:
Relates: #2670


helsner:
This was hell itself IMHO.. the array and property switching was pretty hard and i don't even know how i got it.
Also there was no example in another rule, so i tried to copy it from various places 

BE AWARE: this does only solve one part of the RST-file, as the method call is not changed - should i do that as well in another rule? (would do so, but ask before wasting time)

sabbelasichon:
@helsner I would create another rule for the second part of the RST-File. Really well solved. Nothing to add.

helsner:
OK will do so tomorrow

helsner:
I would merge myself this time if everything is clear hihi

PR #2725, opened 2021-11-04
feat: handle ContentObjectRenderer->getAtagParams()

helsner:
Resolves: #2670

### DIFF
--- a/src/Rector/v11/v5/HandleCObjRendererATagParamsMethodRector.php
+++ b/src/Rector/v11/v5/HandleCObjRendererATagParamsMethodRector.php
@@ -76,11 +76,6 @@ CODE_SAMPLE
         )) {
             return true;
         }
-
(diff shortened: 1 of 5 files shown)


src/Rector/v11/v5/HandleCObjRendererATagParamsMethodRector.php
```diff
@@ -0,0 +1,81 @@
...
+        if ($this->shouldSkip($node)) {
+            return null;
+        }
+
+        if (1 === count($node->args)) {
+            return null;
+        }
+
+        // This might be true or 1, so we are not type strict comparing here
+        if ($this->valueResolver->getValue($node->args[1]->value)) {
```
helsner, line 42: @sabbelasichon 
This i'm not sure about.
In case the value is false the behavior differs. So removing that could cause bugs/site effects.

Issue #2586, opened 2021-09-24
Deprecation: #91806 - BackendUtility viewOnClick


sabbelasichon:
Deprecation: #91806 - BackendUtility viewOnClick

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/11.3/Deprecation-91806-BackendUtilityViewOnClick.html
.. include:: ../../Includes.txt

================================================
Deprecation: #91806 - BackendUtility viewOnClick
================================================

See :issue:`91806`

Description
===========

Method :php:`BackendUtility::viewOnClick()` is discouraged to be used
due to its inline JavaScript and has been deprecated now.

Impact
======

Using the method will trigger a PHP :php:`E_USER_DEPRECATED` error.

Affected Installations
======================

Extensions calling this static method will be affected. The extension
scanner will find usages as strong match.

Migration
=========

The :php:`\TYPO3\CMS\Backend\Routing\PreviewUriBuilder` should be used
instead as described in
:doc:`/Changelog/11.0/Important-91123-AvoidUsingBackendUtilityViewOnClick`.

.. index:: Backend, JavaScript, PHP-API, FullyScanned, ext:backend

PR #2729, opened 2021-11-06
feat: resolve usage of BackendUtility::viewOnClick

helsner:
Resolves: #2586


helsner:
This was a bit more complex for me, but i tried.
I don't really like the chaining of the methods, but could not think of another way.
Also the check for phpstan will fail because i did not manage to resolve the arg "$switchFocus" and provide it properly to the method as array.

I added the output styles, as the result does not fully work, but is a preparation for the further attribute handling.

helsner:
ok, i guess i got it 👌

sabbelasichon:
@helsner Looks promising. What has to be done manually afterwards in which situation? I am not sure if we should already rewrite this one or we should move it to v12.0 or make it optional for those who know about it? Probably both of us ;-)

helsner:
I don't know a 100%. It depends on the individual use case.
But i guess we could make it optional for us and others.

sabbelasichon:
Could you remove it from the config, i am going to ship it asap

sabbelasichon:
I have done it.

PR #2736, opened 2021-11-16
Add stubs to export-ignore

samsonasik:
Hi @sabbelasichon, currently, the longest time to downgrade on `rector` scoping is typo3-rector package. Do you think `stubs` directory needs to be included in the dist? If no, I think it can be registered as export-ignore in `.gitattributes`

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -23,4 +23,4 @@ utils/composer-packages export-ignore
 utils/generator export-ignore
 utils/rules export-ignore
 typo3.constants.php export-ignore
-stubs
+/stubs export-ignore


samsonasik:
PHPStan notice is unrelated

sabbelasichon:
@samsonasik Thank you. No, it is not needed. Once it was exclude than i thought with static reflection it is needed. I was so confused at a certain point and it totally forgot about it.

samsonasik:
@sabbelasichon thank you, could you create new tag (0.11.31 ? ) for it?

PR #2739, opened 2021-11-24
[DX] Make visibility manipulator explicit inject

TomasVotruba:
This is an attempt to make `AbstractRector` slimmer and lighter :) services should be injected only when needed. No just because 10 rules use them.

See Rector core change https://github.com/rectorphp/rector-src/pull/1298


TomasVotruba:
I'm merging so we can update Rector together with this package. It's a BC break as the property visibilty was changed from `protected` to `private`.

PR #2747, opened 2021-12-08
[DX] make use of configure() method

TomasVotruba:
This completes the `configure()` feature update, as in core @rectorphp https://github.com/rectorphp/rector-src/pull/1410

Basically, instead of complex call + constant + inline value object mess:

```php
$services->set(SomeRector::class)
    ->call('configure', [[
        SomeRector::SOME_CONSTANT => ValueObjectInliner::inline([
              new ConfigureValueObject('...');
        ])
    ]]);
```

<br>

To single `configure()` method that handles validation for us:

```php
$services->set(SomeRector::class)
    ->configure([
          new ConfigureValueObject('...');
    ]);
```

:+1:


TomasVotruba:
Ready to :ship:  :slightly_smiling_face: 

If you find any bug here in practical use, let me know. I'll fix it

sabbelasichon:
Amazing. This is so readable now and more intuitive for an unexperienced developer to write the config right. Thank you. Is this part of symfony/dependency-injection now?

TomasVotruba:
> Amazing. This is so readable now and more intuitive for an unexperienced developer to write the config right. 

Thank you! I realized my first idea with constant + value object wrapper is a bit overkill :laughing: This feels more lighter and easier to use. Plus the assertion make sure the input is always right.

> Is this part of symfony/dependency-injection now?

Not really. Symfony actually makes this impossible to implement, as the method is marked `final`. We had to add patch, so the `->set()` return our custom service configurator. That one has `->configure()` method that handles the complexity inside. 

Here is the PR that added it: https://github.com/rectorphp/rector-src/pull/1276/files

sabbelasichon:
Fantastico. I am impressed. DX is such an important feature, but to get it right you have to be an artist like you.

PR #2748, opened 2021-12-08
[phpstan] Improve type resolving without property_exists

TomasVotruba:
Cherry pick of https://github.com/sabbelasichon/typo3-rector/pull/2747

### DIFF
--- a/utils/composer-packages/src/Rector/AddPackageVersionRector.php
+++ b/utils/composer-packages/src/Rector/AddPackageVersionRector.php
@@ -56,29 +56,27 @@ final class AddPackageVersionRector extends AbstractRector
             return null;
         }
 
-        if (! property_exists($node, 'stmts')) {
-            return null;
-        }
+        /** @var Closure $closure */
+        $closure = $node;
 
         /** @var Expression $stmt */
-        foreach ($node->stmts as $stmt) {
+        foreach ($closure->stmts as $stmt) {
             if (! $stmt->expr instanceof Assign) {
                 continue;
             }
 
-            if (! $this->isName($stmt->expr->var, 'composerExtensions')) {
+            $assign = $stmt->expr;
+
+            if (! $this->isName($assign->var, 'composerExtensions')) {
                 continue;
             }
 
-            if (! $stmt->expr->expr instanceof Array_) {
+            if (! $assign->expr instanceof Array_) {
                 continue;
             }
 
-            if (! property_exists($stmt->expr->expr, 'items')) {
-                continue;
-            }
-
-            $stmt->expr->expr->items[] = new ArrayItem(
+            $array = $assign->expr;
+            $array->items[] = new ArrayItem(
                 new New_(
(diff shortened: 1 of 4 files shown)


utils/composer-packages/src/Rector/AddPackageVersionRector.php
```diff
@@ -56,29 +56,27 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        if (! property_exists($node, 'stmts')) {
```
TomasVotruba, line 59: This will actually hide the type to both PhpStorm and PHPStan.

It's better to use instanceof of type check, ideally in strict form. That way both these tools (and Rector :wink: ) can understand context and warn about bugs and refactor :+1:

utils/composer-packages/src/Rector/AddPackageVersionRector.php
```diff
@@ -56,29 +56,27 @@ public function refactor(Node $node): ?Node
...
+            if (! $this->isName($assign->var, 'composerExtensions')) {
                 continue;
             }
 
-            if (! property_exists($stmt->expr->expr, 'items')) {
+            if (! $assign->expr instanceof Array_) {
                 continue;
             }
 
-            $stmt->expr->expr->items[] = new ArrayItem(
```
TomasVotruba, line 81: During testing, I realized that PhpStorm cannot handle multinesting instanceof.

E.g.

```php
if ($node->expr->expr instanceof String_) {
     $node->expr->expr // mixed?
}
```

But can remember one property fetch:

```php
if ($node->expr instanceof String_) {
     $node->expr // String_ of course
}
```

I used this trick here, to make PhpStorm always remember the type :+1:

utils/composer-packages/src/Rector/AddReplacePackageRector.php
```diff
@@ -23,16 +23,19 @@ final class AddReplacePackageRector extends AbstractRector
...
 
-    public function setReplacePackages(array $replacePackages): void
+    /**
+     * @param RenamePackage[] $renamePackages
+     */
+    public function setReplacePackages(array $renamePackages): void
     {
-        $this->replacePackges = $replacePackages;
+        $this->renamePackages = $renamePackages;
     }
```
TomasVotruba, line 39: Just name correction, to match the array of passed value objects :+1:

utils/composer-packages/src/Rector/RemovePackageVersionsRector.php
```diff
@@ -41,29 +41,27 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
-        if (! property_exists($node, 'stmts')) {
-            return null;
-        }
+        /** @var Closure $closure */
+        $closure = $node;
```
TomasVotruba, line 46: Thanks to validation method above, we know that the `$node` is always `Closure` node. Tools cannot handle method calls type check, so we have to remind them the type.

TomasVotruba:
Ready to review and merge :+1:

PR #2749, opened 2021-12-08
Correct RenamePackage name, validate configure() input

TomasVotruba:
Split of https://github.com/sabbelasichon/typo3-rector/pull/2747

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": ">=8.1",
-        "helmich/typo3-typoscript-parser": "^2.3.1",
+        "helmich/typo3-typoscript-parser": "dev-master#0ccb3a6",
         "symfony/var-exporter": "^5.3",
         "symfony/string": "^6.0"
     },
--- a/src/FileProcessor/TypoScript/TypoScriptFileProcessor.php
+++ b/src/FileProcessor/TypoScript/TypoScriptFileProcessor.php
@@ -25,6 +25,7 @@ use Ssch\TYPO3Rector\Contract\FileProcessor\TypoScript\TypoScriptRectorInterface
 use Ssch\TYPO3Rector\Contract\Processor\ConfigurableProcessorInterface;
 use Ssch\TYPO3Rector\FileProcessor\TypoScript\Rector\AbstractTypoScriptRector;
 use Symfony\Component\Console\Output\BufferedOutput;
+use Webmozart\Assert\Assert;
 
 /**
  * @see \Ssch\TYPO3Rector\Tests\FileProcessor\TypoScript\TypoScriptProcessorTest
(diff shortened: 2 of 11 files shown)


utils/composer-packages/src/Command/AddComposerTypo3ExtensionsToConfigCommand.php
```diff
@@ -142,7 +142,7 @@ private function addReplacePackages(ExtensionCollection $collection): void
         $this->decorateNamesToFullyQualified($nodes);
 
         $nodeTraverser = new NodeTraverser();
-        $this->replacePackageRector->setReplacePackages($collection->getReplacePackages());
+        $this->replacePackageRector->configure($collection->getRenamePackages());
```
TomasVotruba, line 145: This rule is configurable, so it should use `configure()` method convention + implement the `Rector\Core\Contract\Rector\ConfigurableRectorInterface` interface

composer.json
```diff
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": ">=8.1",
-        "helmich/typo3-typoscript-parser": "^2.3.1",
+        "helmich/typo3-typoscript-parser": "dev-master#0ccb3a6",
         "symfony/var-exporter": "^5.3",
```
TomasVotruba, line 15: This is just temporary requirement for the PR with fixed Symfony version: https://github.com/martin-helmich/typo3-typoscript-parser/pull/74

In stable package this would be problem, but in Rector we build the final scoped version from dev packages as well, so we're ok :+1: 

As soon the fixed PR is released, we can switch back to stable tag

src/FileProcessor/TypoScript/TypoScriptFileProcessor.php
```diff
@@ -81,9 +82,16 @@ public function getSupportedFileExtensions(): array
...
+     * @param mixed[] $configuration
+     */
     public function configure(array $configuration): void
     {
-        $this->allowedFileExtensions = $configuration[self::ALLOWED_FILE_EXTENSIONS] ?? [];
+        $allowedFileExtensions = $configuration[self::ALLOWED_FILE_EXTENSIONS] ?? $configuration;
+        Assert::isArray($allowedFileExtensions);
+        Assert::allString($allowedFileExtensions);
+
+        $this->allowedFileExtensions = $allowedFileExtensions;
```
TomasVotruba, line 94: It's a good practise to validate user's input - if it's a form, an API or configuration in config.
Here we make sure we fail as soon as `configure()` is called with invalid input and let use know to correct it :+1:

TomasVotruba:
Rebased on `main`

TomasVotruba:
This PR should fix the CI on `main`, as it allow installing Symfony 6

PR #2750, opened 2021-12-08
[phpstan] Resolve PHPStan errors

TomasVotruba:


utils/composer-packages/src/Rector/AddReplacePackageRector.php
```diff
@@ -86,7 +89,7 @@ public function refactor(Node $node): ?Node
             foreach ($this->renamePackages as $renamePackage) {
                 $array->items[] = new ArrayItem(
                     new New_(
-                        new FullyQualified(RenamePackage::class),
+                        new FullyQualified('Rector\Composer\ValueObject\RenamePackage'),
```
TomasVotruba, line 92: This might get accidentaly prefixed during rector/rector-src build and create invalid output, so the value should be `string` type.

TomasVotruba:
Ready to review & merge :+1:

sabbelasichon:
@TomasVotruba Many thanks. Looks fantastic.

Issue #2756, opened 2021-12-13
Docs: Improve docs by adding pagination to next page


helsner:
When using rector i'm missing a pagination on the docs.
I install and see "ok, thats not what i need, i need configuration" and have to go back and forth all the time instead of going with the flow.
Will do myself!

PR #2759, opened 2021-12-14
docs: provide table of contents for each doc segment

helsner:
Resolves: #2756


helsner:
I thought about how to properly do this, with a prev and next button, but to me it is better accessible to have the whole table of contents within each file as it is not that long and navigation is way easier, especially as this is our main documentation.

PR #2766, opened 2021-12-28
[Paralell] Prepare file processors for parallel run

TomasVotruba:
I'm working on parallel run in Rector - https://github.com/rectorphp/rector/issues/2971  :hugs: 

After this PR, https://github.com/rectorphp/rector-src/pull/1583 it is required to return jsonable value objects, so paralell process can merge them around :) 

This PR updates architecture here, so the diff is always returned :+1:

### DIFF
--- a/src/FileProcessor/Yaml/Form/FormYamlFileProcessor.php
+++ b/src/FileProcessor/Yaml/Form/FormYamlFileProcessor.php
@@ -4,7 +4,6 @@ declare(strict_types=1);
 
 namespace Ssch\TYPO3Rector\FileProcessor\Yaml\Form;
 
-use Nette\Utils\Strings;
 use Rector\ChangesReporting\ValueObjectFactory\FileDiffFactory;
 use Rector\Core\Contract\Processor\FileProcessorInterface;
 use Rector\Core\Provider\CurrentFileProvider;
@@ -14,6 +13,7 @@ use Rector\Core\ValueObject\Error\SystemError;
 use Rector\Core\ValueObject\Reporting\FileDiff;
 use Rector\Parallel\ValueObject\Bridge;
 use Ssch\TYPO3Rector\Contract\FileProcessor\Yaml\Form\FormYamlRectorInterface;
+use Ssch\TYPO3Rector\FileProcessor\Yaml\YamlIndentResolver;
 use Symfony\Component\Yaml\Yaml;
 
 /**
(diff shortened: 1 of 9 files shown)


tests/FileProcessor/FlexForms/Fixture/flexforms.xml
```diff
@@ -50,44 +50,43 @@
...
-    </sDEF>
-</sheets>
+                                </items>
+                            </config>
+                        </TCEforms>
+                    </settings.topNewsRestriction>
+                </el>
+            </ROOT>
+        </sDEF>
+    </sheets>
```
TomasVotruba, line 91: Not sure why, but the Rector change fixed wrong indent here :)

src/FileProcessor/Yaml/Form/FormYamlFileProcessor.php
```diff
@@ -54,15 +65,27 @@ public function process(File $file, Configuration $configuration): void
...
         if ($newYaml === $yaml) {
-            return;
+            return $systemErrorsAndFileDiffs;
         }
 
-        $newFileContent = Yaml::dump($newYaml, 99);
+        $spaceCount = $this->yamlIndentResolver->resolveIndentSpaceCount($oldYamlContent);
+
+        $newFileContent = Yaml::dump($newYaml, 99, $spaceCount);
         $file->changeFileContent($newFileContent);
```
TomasVotruba, line 74: Before, the YAML was printed with default 2 space indent. That broke the tabs spacing and returned false positive.

The `YamlIndentResolver` service now makes sure the original indent is always kept. The regular expression might need some tweaks, but the main idea is finished :slightly_smiling_face:

TomasVotruba, line 74: Before, the YAML was printed with default 2 space indent. That broke the tabs spacing and returned false positive.

The `YamlIndentResolver` service now makes sure the original indent is always kept. The regular expression might need some tweaks, but the main idea is finished :slightly_smiling_face:

TomasVotruba:
I'm mergin, as rector-src depends on packages to be with same parallel paradigm.
Do not relase new version yet, just in case :)  we need to test the main feature first.

src/FileProcessor/Yaml/Form/FormYamlFileProcessor.php
staabm: alternatively/additoinally the dumper could try to read the `.editorconfig` to get the expected whitespace config

sabbelasichon: Thanks a lot men. Looks promising.

PR #2792, opened 2022-01-13
Fix configure calls in configuration files

sabbelasichon:
Resolves: #2791


TomasVotruba:
Thank you :+1: 

I've missed those cases during refactoring, my bad.

sabbelasichon:
No problem. Thank you.

Issue #2724, opened 2021-11-03
Add setterInjection to constructorInjection rector


sabbelasichon:
https://gist.github.com/alexanderschnitzler/f664b458f2bc47833c7a55f93b5a363d

sabbelasichon:
@helsner Would you like to create a rule with tests for that? Should be an optional rule not part of a specific TYPO3 version.

helsner:
will do someday 👍

PR #2793, opened 2022-01-15
feat: provide general inject replacement rector

helsner:
Resolves: #2724


helsner:
@sabbelasichon is simply copied the gist of the issue and added the necessary test segments, which could be extended if required?
Other than that i'm wondering whats wrong with phpstan..

sabbelasichon:
@helsner I am going to fix the phpstan errors. Everything else looks good.

PR #2795, opened 2022-01-19
Add tests to check against merge of TCA config

froemken:
In TYPO3 9 and below there was an ArrayUtility::mergeRecursiveWithOverrule in Extbase:

```
if (isset($currentClassSettings['mapping']['columns']) && is_array($currentClassSettings['mapping']['columns'])) {
                        \TYPO3\CMS\Core\Utility\ArrayUtility::mergeRecursiveWithOverrule($columnMapping, $currentClassSettings['mapping']['columns'], true, false);
                    }
```

which was often used to individually override TCA by page.

This array merge was completely removed with TYPO3 10. So nothing of all these specials should be migrated into Classes.php

### DIFF
--- a/tests/FileProcessor/TypoScript/Expected/Extbase.php.inc
+++ b/tests/FileProcessor/TypoScript/Expected/Extbase.php.inc
@@ -50,6 +50,9 @@ return [
             ],
         ],
     ],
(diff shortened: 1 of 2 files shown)


sabbelasichon:
Thank you. Resolved by #2802

PR #2797, opened 2022-01-21
Place Classes.php from ExtbasePersistenceTypoScriptRector in nested d…

AMartinNo1:
…irectory


sabbelasichon:
@AMartinNo1 Thanks, but this does not work. Imagine all your extensions lying in packages/ext/foo packages/ext/bar than the Classes.php would be created above the actual extension folder. I will come up with a dynamic solution to identify where the ext_emconf.php is lyring around.

AMartinNo1:
> @AMartinNo1 Thanks, but this does not work. Imagine all your extensions lying in packages/ext/foo packages/ext/bar than the Classes.php would be created above the actual extension folder. I will come up with a dynamic solution to identify where the ext_emconf.php is lyring around.

Alright. Sounds good 😊

PR #2798, opened 2022-01-21
Allow to specify target extension in ExtbasePersistenceTypoScriptRector

AMartinNo1:
Example config:

```php
    $pathToExtension = __DIR__ . '/packages/my-extension/';
    $parameters->set(Option::PATHS, [$pathToExtension]);
    $services = $containerConfigurator->services();
    $services->set(ExtbasePersistenceTypoScriptRector::class)->configure(['target' => $pathToExtension]);
```

### DIFF
--- a/src/FileProcessor/TypoScript/Rector/ExtbasePersistenceTypoScriptRector.php
+++ b/src/FileProcessor/TypoScript/Rector/ExtbasePersistenceTypoScriptRector.php
@@ -42,9 +42,15 @@ final class ExtbasePersistenceTypoScriptRector extends AbstractTypoScriptRector 
 
     private SmartFileInfo $fileTemplate;
 
+    /**
+     * @var array{target: string}
+     */
+    private array $configuration;
+
     public function __construct(TemplateFinder $templateFinder)
     {
-        $this->filename = getcwd() . '/Configuration_Extbase_Persistence_Classes.php';
+        $this->configuration['target'] = getcwd();
+        $this->filename = '/Configuration_Extbase_Persistence_Classes.php';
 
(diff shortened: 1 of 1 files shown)


src/FileProcessor/TypoScript/Rector/ExtbasePersistenceTypoScriptRector.php
```diff
@@ -126,6 +133,7 @@ public function getMessage(): string
 
     public function configure(array $configuration): void
     {
+        $this->configuration = $configuration;
```
AMartinNo1, line 136: If we can access whatever is defined with `$parameters->set(Option::PATHS, [$pathToExtension]);`, I guess this would be preferrable. But for now, this is my suggestion.

src/FileProcessor/TypoScript/Rector/ExtbasePersistenceTypoScriptRector.php
```diff
@@ -116,7 +122,8 @@ public function convert(): ?AddedFileWithContent
             return sprintf('\%s::class', $string);
         });
 
-        return new AddedFileWithContent($this->filename, $content);
+        $fullPath = rtrim($this->configuration['target'], '/') . '/' . ltrim($this->filename, '/');
```
AMartinNo1, line 125: It's tedious to check whether a trailing or leading slash is required. So, with this it does not matter what you provide in the config - get the upgrade you're working on done, w/o bothering about slashes.

sabbelasichon:
Many thanks for the idea. Really good idea. I am gonna fix this via a different PR.

Issue #2800, opened 2022-01-23
BackendUtility::getModTSconfig


ayacoo:
Source
`$configArray = BackendUtility::getModTSconfig($pid, 'mod.web_list');`

Target
`$configArray['properties'] = BackendUtility::getPagesTSconfig($pid)['mod.']['web_list.'] ?? [];`

Changelog via https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.3/Deprecation-84993-DeprecateSomeTSconfigRelatedMethods.html

PR #2801, opened 2022-01-23
[DRAFT][FEATURE] Add BackendUtilityGetModTsConfigRector

ayacoo:
Related #2800

There are still some things that are unclear to me here. I have tried to work on the things.

Unknown are the following things:
- A multiple ArrayDimFetch
- A merge of a function with an ArrayDimFetch

Unfortunately, I could not remove the original node. I had switched to Stmt/Expression here, but then my code location was also simply no longer recognized.

Maybe you can give me 1-2 tips where to get the knowledge. I would like to work on it here ;-)


sabbelasichon:
@ayacoo Thank you. Will have a look.

Issue #2794, opened 2022-01-19
Do not try to migrate TypoScript mapping-parts as they were completely removed


froemken:
# Bug Report

| Subject        | Details                 |
| :------------- | :--------------|
| Rector version | 0.12.12             |

With TYPO3 ChangeLog 87623 the sub-property `mapping` was removed completely: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/10.0/Breaking-87623-ReplaceConfigpersistenceclassesTyposcriptConfiguration.html
But rector-typo3 tries to migrate the `mapping` part into Classes.php:
```
return [
    \JWeiland\Maps2\Domain\Model\PoiCollection::class => [
        'properties' => [
            'passthrough' => [
                'fieldName' => 'distance',
            ],
        ],
    ],
];
```

`passthrough` is totally wrong here. `passthrough` is the TCA type of `distance`, but was used here as columnName.

## Minimal PHP Code Causing Issue

_sorry, can't use demo page, as it needs PHP to convert_

**TypoScript**

```
plugin.tx_maps2 {
  persistence {
    classes {
      JWeiland\Maps2\Domain\Model\PoiCollection {
        mapping.columns {
          distance.config {
            type = passthrough
          }
        }
      }
    }
  }
}
```

**rector.php**
```
<?php

use Symfony\Component\DependencyInjection\Loader\Configurator\ContainerConfigurator;

return static function (ContainerConfigurator $containerConfigurator): void {

    $containerConfigurator->import(\Ssch\TYPO3Rector\Set\Typo3SetList::TYPO3_87);
    $containerConfigurator->import(\Ssch\TYPO3Rector\Set\Typo3SetList::TYPO3_95);
    $containerConfigurator->import(\Ssch\TYPO3Rector\Set\Typo3SetList::TYPO3_104);
};
```

## Expected Behaviour

Do not try to migrate any kind of TypoScript `mapping`

PR #2802, opened 2022-01-25
Do not use VarExporter anymore

sabbelasichon:
@TomasVotruba Could get rid of VarExporter by reformatting after prettyPrinting. So one dependency less.

Resolves: #2794

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,6 @@
     "require": {
         "php": ">=8.1",
         "helmich/typo3-typoscript-parser": "dev-master#0ccb3a6",
-        "symfony/var-exporter": "^6.0",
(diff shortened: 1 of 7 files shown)


TomasVotruba:
That's awesome :partying_face:  thank you for taking the time and effort :hugs:

TomasVotruba:
The results in compiler Rector? 1150 lines less to maintain :+1: 
https://github.com/rectorphp/rector/commit/fc1e56bf9f12636fb9e849d7f6af42f091f9e87c#diff-5ae063baecc4813da5c3f8e177d883ea93596a4c6abc8e5a2eaad7204923ed5f

PR #2814, opened 2022-02-20
[update] Use stable Hemlich with Symfony 6 support

TomasVotruba:
- bump to stable gemlich, since no Symfony 6 is supported
- bump helmich deps to Symfony 6 compatible one

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -11,7 +11,7 @@
     ],
     "require": {
         "php": ">=8.1",
-        "helmich/typo3-typoscript-parser": "dev-master#0ccb3a6",
+        "helmich/typo3-typoscript-parser": "^2.4.1",
         "symfony/string": "^6.0"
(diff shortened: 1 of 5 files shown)


composer.json
```diff
@@ -11,17 +11,17 @@
     ],
     "require": {
         "php": ">=8.1",
-        "helmich/typo3-typoscript-parser": "dev-master#0ccb3a6",
+        "helmich/typo3-typoscript-parser": "2.4.1 as dev-master",
```
TomasVotruba, line 14: Temporary, until similar PR gets merged to Rector. Now there is circular reference preving to upgrade this package, this is how to make composer update pass.

TomasVotruba:
I'm mergin it, so we can upgrade core rector-src to use new helmich package :+1:

PR #2818, opened 2022-02-28
Use $this->nodesToAddCollector-> call directly for addNodes?(?:Before|After)Node call

samsonasik:
To be able to remove deprecated methods addNodes?(?:Before|After)Node on AbstractRector, ref https://github.com/rectorphp/rector-src/pull/1879


samsonasik:
All checks have passed 🎉 @sabbelasichon @TomasVotruba it is ready for review.

TomasVotruba:
Thank you :+1:

PR #2832, opened 2022-04-09
Use custom NodePrinterInterface

TomasVotruba:
See https://github.com/rectorphp/rector-src/pull/2036


sabbelasichon:
@TomasVotruba Thank you very much.

PR #2837, opened 2022-04-20
Make use of RectorConfig and ECSConfig

TomasVotruba:
Since 0.12.21 it is possible to use lighter RectorConfig class, that contains full method autocomplete and validation.
This way we have less dependency on Symfony and easier setup :+1:


TomasVotruba:
Good to go :)

PR #2840, opened 2022-04-20
Make use of RectorConfig and ECSConfig

TomasVotruba:


TomasVotruba:
Phew :) the huge configs makes my PhpStorm lag, but Rector helped itself to upgrade :)

I hope I got everything right.  This is good to go :+1: 

Could you check?

TomasVotruba:
💪

simonschaufi:
@TomasVotruba How is the file `ssch/typo3-rector/config/composer/typo3-104-composer-packages-extensions.php` generated? It has an error: 

```
[ERROR] Call to undefined method "Rector\Config\RectorConfig::ruleWithConfigure()" in                                  
         /xxx/vendor/rector/rector/vendor/ssch/typo3-rector/src/Set/../../config/composer/typo3-10
         4-composer-packages-extensions.php (which is being imported from "/xxx/rector.php").  
```

TomasVotruba:
@simonschaufi Thanks for reporting :+1: 

I don't know, but when you look for "ruleWithConfigure" string, it should be clear.
Could you send PR to fix it?

simonschaufi:
@TomasVotruba I found it only within that file but that file looks like it's generated as it contains plenty of versions of TYPO3 extensions (which is not up-to-date anyway) and should be generated again.

TomasVotruba:
I'd simply replace it, as it didn't change last month.

simonschaufi:
@TomasVotruba Sorry that this turns into a support chat but I found the command that generates the file but when calling it, I get the error from rector-src: " Node "PhpParser\Node\Expr\Closure" with parent of "PhpParser\Node\Stmt\Return_" is missing scope required for scope refresh."

The command is: `vendor/bin/rector add-composer-typo3-extensions-to-config`

I'm totally lost what "missing scope" means here.

TomasVotruba:
This will be tricky... is the new `Closure` node somehow traversed?

simonschaufi:
I have no idea about that

TomasVotruba:
It looks like missused of `AddPackageVersionRector` for simple node traverser

TomasVotruba:
I'm working on refactoring here https://github.com/sabbelasichon/typo3-rector/pull/2998

It's almost done, just super slow. I'll seen during weekend, what can be improved :) thanks for reporting :+1:

TomasVotruba:
@simonschaufi So after 5 hours I've come to few findings :) the command is super slow because:

* it fetches all typo3 packages from Packagist, one by one... in total 2777 API calls
* it parses and loads files in `/config/composer`, that have thousands of lines of nested method calls in PHP, so it takes Rector long time to load and process
* it generates probably hundreds such configuration items, based on dependencies of 2777 packagist

I see this code first time, as I only improve changes from Rector core here, so bear with me. 

The idea to improve this I had was:

* do not work with PHP, but only dump json package renames
* load those package renames via foreach() in sets

I'm not sure this is helpful, as the command takes ~30 mins to run (maybe more) and any break is hell to debug. 

Another idea is:

* drop this component of this package, because it seems too heavy to maintain and work with. I guess 90 % of changes takes the most churn
 

Saying that, I don't want to invest more energy anymore, as to make it useful would take probably 3-5 days, as everything is new to me here. 

But you can continue with the exploration. Here is some work you can continue with https://github.com/sabbelasichon/typo3-rector/pull/2998/files

sabbelasichon:
Thanks Tomas. It worked quite well. Don't why it is not anymore. Lost track of it. I think we should drop this as it turned out not to be used that much. I will have a deeper look next week.

simonschaufi:
@TomasVotruba thank you so much for your investigation!
@sabbelasichon I noticed that there has been a GitHub action to update the composer package list but this is also gone, at least I found some commits like: "[automated] Update composer packages (#2427)"

PR #2967, opened 2022-05-10
Refactor AddIconsToReturnRector, AddArgumentToSymfonyCommandRector and AddCommandsToReturnRector to proper node factories

TomasVotruba:
Like controller cannot include another controller, the rule cannot use another rule for refactoring. Things can go wrong :)
The correct way would be factory or simple node traverser :+1: 

Discoverd during https://github.com/rectorphp/rector-src/pull/2239


src/Rector/v11/v5/RegisterIconToIconFileRector.php
```diff
@@ -164,26 +163,24 @@ private function addNewIconToIconsFile(
...
+
         if (is_string($existingIcons)) {
             $stmts = $this->simplePhpParser->parseString($existingIcons);
+            $this->traverseNodesWithCallable($stmts, function (Node $node) use ($iconArrayItem) {
+                if (! $node instanceof Array_) {
+                    return null;
+                }
+
+                $node->items[] = $iconArrayItem;
+                return NodeTraverser::DONT_TRAVERSE_CHILDREN;
```
TomasVotruba, line 176: Add it just once, avoids infinity loop

TomasVotruba:
Thanks for fast merge 👍

Could you verify it works as expected? 😆 I think the tests only verify existence of file, not exact contents.

sabbelasichon:
The tests are also checking the contents:
https://github.com/sabbelasichon/typo3-rector/blob/main/tests/Rector/v9/v5/ExtbaseCommandControllerToSymfonyCommandRector/ExtbaseCommandControllerToSymfonyCommandRectorTest.php#L28

https://github.com/sabbelasichon/typo3-rector/blob/main/tests/Rector/v11/v5/RegisterIconToIconFileRector/RegisterIconToIconFileRectorTest.php#L29

Well done.

TomasVotruba:
Ah, that's great. I was almost sleeping at 2 AM, so I was just surprised test were passing so easily :laughing: 


Thanks :+1:

Issue #2947, opened 2022-05-06
Deprecation: #97217 - Moved TypoLinkCodecService to EXT:core


sabbelasichon:
Deprecation: #97217 - Moved TypoLinkCodecService to EXT:core

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-97217-MovedTypoLinkCodecServiceToEXTcore.html
.. include:: /Includes.rst.txt

============================================================
Deprecation: #97217 - Moved TypoLinkCodecService to EXT:core
============================================================

See :issue:`97217`

Description
===========

The :php:`TypoLinkCodecService` class is used to encode and decode link
parameters, which are usually next to the actual link the `target` or
`class` information. This functionality is not directly bound to frontend
specific logic. To resolve cross dependencies the class has been moved to
the `LinkHandling` namespace in EXT:core. The old namespace has therefore
been deprecated.

Impact
======

The namespace has changed from :php:`\TYPO3\CMS\Frontend\Service\TypoLinkCodecService`
to :php:`\TYPO3\CMS\Core\LinkHandling\TypoLinkCodecService` and the old namespace
has been deprecated.

Affected Installations
======================

All installations using the deprecated namespace
:php:`\TYPO3\CMS\Frontend\Service\TypoLinkCodecService`. The extension
scanner will report usages.

Migration
=========

Replace usages with the new namespace
:php:`\TYPO3\CMS\Core\LinkHandling\TypoLinkCodecService`
in custom extension code.

.. index:: PHP-API, FullyScanned, ext:frontend

helsner:
i will do that or give that to my colleague for practice!

PR #2968, opened 2022-05-11
feat: add namespace migration for TypoLinkCodecService

dominikdoerr:
Resolves: #2947


sabbelasichon:
Thanks a lot for your contribution. I must admit that i already have some generator in place for those Class-Alias-Maps.
It is a local utility repo i have. I will add those immediately.

Issue #2945, opened 2022-05-06
Deprecation: #97109 - TCA type none "cols" option


sabbelasichon:
Deprecation: #97109 - TCA type none "cols" option

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-97109-TCATypeNoneColsOption.html
.. include:: /Includes.rst.txt

=================================================
Deprecation: #97109 - TCA type none "cols" option
=================================================

See :issue:`97109`

Description
===========

The TCA type `none` had two option keys for the same functionality: `cols` and
`size`. In order to simplify the available configuration, `cols` has been
dropped in favour of `size`.

Impact
======

Defining the option `cols` for the TCA type `none` will log a deprecation
message. An automatic migration is in place, which will be displayed in the TCA
Migrations view of the Upgrade module.

Affected Installations
======================

All installations using the `cols` option for the TCA type `none`.

Migration
=========

Rename the option `cols` to `size`.

Before:

.. code-block:: php

    'columns' => [
        'aColumn' => [
            'config' => [
                'type' => 'none',
                'cols' => 20,
            ],
        ],
    ],

After:

.. code-block:: php

    'columns' => [
        'aColumn' => [
            'config' => [
                'type' => 'none',
                'size' => 20,
            ],
        ],
    ],

.. index:: TCA, NotScanned, ext:backend

PR #2970, opened 2022-05-11
feat: migrate cols to size for TCA type none

dominikdoerr:
Resolves: #2945


sabbelasichon:
@dominikdoerr Could you also create a fixture where the type is not "none" and no change is done accidentally.

dominikdoerr:
@sabbelasichon I added another column to the fixture.

Issue #2954, opened 2022-05-06
Deprecation: #97384 - TCA option "eval=null" replaced with "nullable"


sabbelasichon:
Deprecation: #97384 - TCA option "eval=null" replaced with "nullable"

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-97384-TCAOptionNullable.html
.. include:: /Includes.rst.txt

=====================================================================
Deprecation: #97384 - TCA option "eval=null" replaced with "nullable"
=====================================================================

See :issue:`97384`

Description
===========

The TCA option :php:`eval=null` has been replaced with the boolean option
:php:`nullable`.

Impact
======

The TCA option :php:`eval=null` will be automatically migrated to
:php:`'nullable' => true`. The migration will trigger a deprecation level log
entry.

Affected Installations
======================

All installations defining the :php:`null` value in their :php:`eval` list.

Migration
=========

To migrate your TCA add the TCA option :php:`'nullable' => true` and remove the
:php:`null` value from the field's :php:`eval` list.

.. code-block:: php

    // Before

    'columns' => [
       'nullable_column' => [
           'title' => 'A nullable field',
           'config' => [
               'eval' => 'null',
           ],
       ],
    ],

    // After

    'columns' => [
       'nullable_column' => [
           'title' => 'A nullable field',
           'config' => [
               'nullable' => true,
           ],
       ],
    ],

.. index:: TCA, NotScanned, ext:backend

simonschaufi:
I'm on it

simonschaufi:
duplicates #3564

PR #2979, opened 2022-05-21
[FEATURE] Add TCA option nullable instead of eval=null

simonschaufi:
See https://review.typo3.org/c/Packages/TYPO3.CMS/+/74319/5/typo3/sysext/core/Classes/Migrations/TcaMigration.php

Resolves: #2954

### DIFF
--- a/config/v12/tca-120.php
+++ b/config/v12/tca-120.php
@@ -1,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+use Rector\Config\RectorConfig;
+use Ssch\TYPO3Rector\Rector\v12\v0\MigrateNullFlagRector;
+
+return static function (RectorConfig $rectorConfig): void {
+    $rectorConfig->import(__DIR__ . '/../config.php');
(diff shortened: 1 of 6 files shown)


src/Helper/TcaHelperTrait.php
```diff
@@ -113,6 +113,65 @@ protected function hasKeyValuePair(Array_ $configValueArray, string $configKey,
...
+     *
+     * @param non-empty-string $delim Delimiter string to explode with
+     * @param string $string The string to explode
+     * @param bool $removeEmptyValues If set, all empty values will be removed in output
+     * @param int $limit If limit is set and positive, the returned array will contain a maximum of limit elements with
+     * the last element containing the rest of string. If the limit parameter is negative, all components
+     * except the last -limit are returned.
+     * @return string[] Exploded values
+     */
+    protected static function trimExplode(
```
sabbelasichon, line 144: This method already exists in the ArrayUtility class. Could you use it from there.

sabbelasichon:
@simonschaufi Thanks a lot. Looks very good

PR #2994, opened 2022-05-26
[BUGFIX] Add type as namespace

simonschaufi:


simonschaufi:
I have overlooked that the file is put in a subdirectory depending on the rector type, but then the namespace doesn't match to the folder. Of course, it would be better to have all rectors in the same folder, but then I would need to do some more magic and remove the "type directory". I also saw that TypoScript rectors are also put into `src/FileProcessor/TypoScript/Rector/v12/v0` but then we need some kind of mapping where to put each rector type.

TomasVotruba:
Hi, this looks good to me.

Could you rebase on `main`?

TomasVotruba (approved):

TomasVotruba:
👍

simonschaufi:
@TomasVotruba please use rebase mode to merge, to get a nice and clean git history without any merge commits. Thanx.

TomasVotruba:
I take the default one and prefer it.

PR #2996, opened 2022-05-27
[DX] Remove TypoScriptFileProcessor dependency on editorconfig, rather use the current file spacing setup

TomasVotruba:
I'm cleaning Rector from a code-formatting, to make it less widespread :) and better focused on refactoring itself

Ref https://github.com/rectorphp/rector-src/pull/2378

### DIFF
--- a/src/FileProcessor/TypoScript/TypoScriptFileProcessor.php
+++ b/src/FileProcessor/TypoScript/TypoScriptFileProcessor.php
@@ -13,6 +13,7 @@ use Helmich\TypoScriptParser\Parser\Printer\PrettyPrinterConfiguration;
 use Helmich\TypoScriptParser\Parser\Traverser\Traverser;
 use Helmich\TypoScriptParser\Parser\Traverser\Visitor;
 use Helmich\TypoScriptParser\Tokenizer\TokenizerException;
+use Nette\Utils\Strings;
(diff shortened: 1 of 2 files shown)


src/FileProcessor/TypoScript/TypoScriptFileProcessor.php
```diff
@@ -143,36 +141,28 @@ private function processFile(File $file): void
...
             $this->typoscriptPrinter->setPrettyPrinterConfiguration($prettyPrinterConfiguration);
 
             $printStatements = $this->filterRemovedStatements($originalStatements, $file);
-
             $this->typoscriptPrinter->printStatements($printStatements, $this->output);
 
             $newTypoScriptContent = $this->applyTypoScriptPostRectors($this->output->fetch());
-
-            $typoScriptContent = rtrim($newTypoScriptContent) . $editorConfiguration->getNewLine();
+            $typoScriptContent = rtrim($newTypoScriptContent) . "\n";
```
TomasVotruba, line 165: This might require better endline resolution in the future. But having the most tools, even php-parser this endline as hardcoded default, I think we're ok with json as well: https://github.com/nikic/PHP-Parser/blob/678ccbe0720549c51d30b76d094a117eac819e9e/lib/PhpParser/PrettyPrinterAbstract.php#L159

PR #2998, opened 2022-05-28
[CI] Make use of simple node traverser to add nodes

TomasVotruba:
Ref https://github.com/sabbelasichon/typo3-rector/pull/2840#issuecomment-1140104092


.github/workflows/code_analysis.yaml
```diff
@@ -40,6 +40,10 @@ jobs:
                         name: 'Check Docs'
                         run: composer docs --ansi
 
+                    -
+                        name: 'Composer Config Generator'
+                        run: vendor/bin/rector typo3-extensions --ansi
```
TomasVotruba, line 45: I took me a while to figure out, what this is :)

As far as I understand:

* it loads all typo3 extension from packagist
* it resolves all their versions
* it generates config for each of them to each typo3-related composer configs
* it prints them back

There are 2 important rules on added and changed composer package version. Those can be Rector rules. 

The `typo3-extensions` command basically parses files, runs Rector rules manually and prints the content back. It's 1:1 to `process` command, just the configuration is dynamic and outside :) 

I think it would be better to use `process` command architecture, as it is obvious to use in Rector and does not break on internal Rector changes. Thus problems like https://github.com/sabbelasichon/typo3-rector/pull/2840#issuecomment-1140104092

Then we can write simple Rector rules and work with them with Rector API without maintaining Rector clone :)

TomasVotruba:
Ref https://github.com/sabbelasichon/typo3-rector/pull/2840#issuecomment-1140252617

PR #3000, opened 2022-05-28
[DX] Make command naming more explicit

TomasVotruba:
- [DX] simplify command naming
- replace deprecated Broker to ReflectionProvider in PHPStanp

### DIFF
--- a/utils/composer-packages/src/Command/AddComposerTypo3ExtensionsToConfigCommand.php
+++ b/utils/composer-packages/src/Command/AddComposerTypo3ExtensionsToConfigCommand.php
@@ -23,7 +23,6 @@ use Symfony\Component\Console\Command\Command;
 use Symfony\Component\Console\Helper\ProgressBar;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symplify\PackageBuilder\Console\Command\CommandNaming;
 use Symplify\SmartFileSystem\SmartFileInfo;
 use Symplify\SmartFileSystem\SmartFileSystem;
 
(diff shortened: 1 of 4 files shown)


utils/composer-packages/src/Command/AddComposerTypo3ExtensionsToConfigCommand.php
```diff
@@ -44,8 +43,7 @@ public function __construct(
 
     protected function configure(): void
     {
-        $this->setName(CommandNaming::classToName(self::class));
-        $this->setAliases(['typo3-extensions']);
+        $this->setName('typo3-extensions');
         $this->setDescription('[DEV] Add TYPO3 extensions from packagist.org to composer configuration');
```
TomasVotruba, line 47: At first it seemed like a great idea, but I learned the command naming is now too magic and hard to read in the code.
The `typo3-extensions` is much more clear :)

utils/generator/src/Command/Typo3GenerateCommand.php
```diff
@@ -42,8 +41,8 @@ public function __construct(
 
     protected function configure(): void
     {
-        $this->setName(CommandNaming::classToName(self::class));
-        $this->setAliases(['typo3-create', 'typo3-rector']);
+        $this->setName('typo3-create');
+        $this->setAliases(['typo3-create']);
         $this->setDescription('[DEV] Create a new TYPO3 Rector, in a proper location, with new tests');
```
TomasVotruba, line 46: I've checked the docs and  "typo3-rector" is always used for the name of this project. 
So I think the "create" or "generate" would be better name.

utils/generator/src/Command/Typo3GenerateCommand.php
```diff
@@ -42,8 +41,8 @@ public function __construct(
 
     protected function configure(): void
     {
-        $this->setName(CommandNaming::classToName(self::class));
-        $this->setAliases(['typo3-create', 'typo3-rector']);
+        $this->setName('typo3-generate');
+        $this->setAliases(['typo3-create']);
         $this->setDescription('[DEV] Create a new TYPO3 Rector, in a proper location, with new tests');
```
TomasVotruba, line 46: I've checked the docs and "typo3-rector" is always used for the name of this project.
So I think the "create" or "generate" would be better name.

PR #3001, opened 2022-05-28
[DX] typo3-extensions command cleanup

TomasVotruba:


utils/composer-packages/src/Command/AddComposerTypo3ExtensionsToConfigCommand.php
```diff
@@ -50,12 +50,12 @@ protected function configure(): void
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
         $typo3Versions = $this->createTypo3Versions();
-        $packages = $this->packageResolver->findAllPackagesByType('typo3-cms-extension');
+        $packages = $this->packagistPackageResolver->findAllPackagesByType('typo3-cms-extension');
 
         $progressBar = new ProgressBar($output, count($packages));
```
TomasVotruba, line 55: Took me a while to figure out all data are coming prom Packagist API :)

utils/composer-packages/src/PackageResolver.php
```diff
@@ -1,18 +0,0 @@
...
-
-interface PackageResolver
-{
-    public function findPackage(ComposerPackage $package): ExtensionCollection;
-
-    /**
-     * @return ComposerPackage[]
-     */
-    public function findAllPackagesByType(string $type): array;
-}
```
TomasVotruba, line 18: The contract here is only one-use. It does not allow to mock ;) nor another implementation is expected. So we can use the service directly with narrowing the scope :+1:

PR #3004, opened 2022-05-28
[TASK] Move rectors into correct TYPO3 version namespace

simonschaufi:


sabbelasichon:
@simonschaufi Really nice, thank you.

simonschaufi:
@sabbelasichon you're welcome but this might be breaking for some pople who import the rules directly and not via the set but hey, we are not yet v1.0 :stuck_out_tongue:

Issue #2875, opened 2022-05-06
Breaking: #96351 - Unused TemplateService->updateRootlineData method removed


sabbelasichon:
Breaking: #96351 - Unused TemplateService->updateRootlineData method removed

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96351-UnusedTemplateService-updateRootlineDataMethodRemoved.html
.. include:: /Includes.rst.txt

============================================================================
Breaking: #96351 - Unused TemplateService->updateRootlineData method removed
============================================================================

See :issue:`96351`

Description
===========

The PHP method `TemplateService->updateRootlineData()` is removed.

It was used as a workaround to update the fetched Rootline with
translated pages until TYPO3 v10. This was necessary because the
TypoScript information contained the language information, and
then the page translations were loaded accordingly.

Since TYPO3 v11 the language is resolved earlier, at the same
time as the page ID, and the mechanism became obsolete.

Impact
======

Calling the method in PHP will throw a fatal PHP error, as the method does not exist anymore.

Affected Installations
======================

TYPO3 installations, mainly legacy installations with legacy
extensions using this method to boot up their own TypoScript
parsing.

Migration
=========

Calling this method is not needed anymore and can be removed
from the affected code.

.. index:: Frontend, PHP-API, FullyScanned, ext:core

PR #3008, opened 2022-05-29
[FEATURE] Remove TemplateService->updateRootlineData()

helsner:
Resolves: #2875

### DIFF
--- a/src/Rector/v12/v0/typo3/RemoveUpdateRootlineDataRector.php
+++ b/src/Rector/v12/v0/typo3/RemoveUpdateRootlineDataRector.php
@@ -1,0 +1,65 @@
+<?php
+
+declare(strict_types=1);
+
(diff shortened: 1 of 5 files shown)


src/Rector/v12/v0/typo3/RemoveUpdateRootlineDataRector.php
```diff
@@ -0,0 +1,65 @@
...
+    {
+        return [MethodCall::class];
+    }
+
+    /**
+     * @param Node\Expr\MethodCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if (! $this->nodeTypeResolver->isMethodStaticCallOrClassMethodObjectType(
```
helsner, line 33: this nodeTypeResolver call seems wrong to me at this place, will check again

Issue #2931, opened 2022-05-06
Deprecation: #96500 - ContentObjectRenderer->getMailTo


sabbelasichon:
Deprecation: #96500 - ContentObjectRenderer->getMailTo

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-96500-ContentObjectRenderer-getMailTo.html
.. include:: /Includes.rst.txt

======================================================
Deprecation: #96500 - ContentObjectRenderer->getMailTo
======================================================

See :issue:`96500`

Description
===========

Since :issue:`96483`, the :html:`<f:link.email/>` ViewHelper is directly
using `TypoLink` for the email link generation. As a result, the
:php:`ContentObjectRenderer->getMailTo()` method was only used in
:php:`EmailLinkBuilder`, the central place for building email links
with `TypoLink`.

To stick to the separation of concerns principle, the corresponding
functionality has been moved to :php:`EmailLinkBuilder->processEmailLink()`
and the :php:`ContentObjectRenderer->getMailTo()` method has been
marked as deprecated.

Impact
======

Calling :php:`ContentObjectRenderer->getMailTo()` will trigger a
PHP :php:`E_USER_DEPRECATED` error. The extension scanner will
find usages as weak match.

Affected Installations
======================

All installations directly calling :php:`ContentObjectRenderer->getMailTo()`
in custom extension code.

Migration
=========

All occurences in extension code have to be replaced by
:php:`EmailLinkBuilder->processEmailLink()`.

.. index:: Frontend, PHP-API, FullyScanned, ext:frontend

sabbelasichon:
https://github.com/TYPO3/typo3/blob/main/typo3/sysext/frontend/Classes/ContentObject/ContentObjectRenderer.php#L4663

helsner:
hi @sabbelasichon 
i need your opinion on this one..
The replacement should call the construct of EmailLinkBuilder, which has one mandatory and one optional argument, while the optional is the TSFE.
In your link the replacement is `$cObj->getTypoScriptFrontendController()` - but this method is marked as @internal
public, but internal with the message: `* @internal this is set to public so extensions such as EXT:solr can use the method in tests.`

We could leave that out or add it? i dont know best practice here

PR #3009, opened 2022-05-29
feat: replace cObj->getMailTo() with EmailLinkBuilder

helsner:
Resolves: #2931


src/Rector/v12/v0/typo3/UseEmailLinkBuilderRector.php
```diff
@@ -0,0 +1,91 @@
...
+     * @param Node\Expr\MethodCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if ($this->shouldSkip($node)) {
+            return null;
+        }
+
+        $arguments = [
+            // $node->args[0] ?? new String_(''),
```
sabbelasichon, line 39: Could you remove the comments here

stubs/TYPO3/CMS/Frontend/Typolink/EmailLinkBuilder.php
```diff
@@ -0,0 +1,18 @@
+<?php
+
+declare(strict_types=1);
+
+
+namespace TYPO3\CMS\Frontend\Typolink\EmailLinkBuilder;
```
sabbelasichon, line 6: Could you add the guard clause at the beginning

```php
if (class_exists('TYPO3\CMS\Frontend\Typolink\EmailLinkBuilder')) {
    return;
}

Comment (sabbelasichon) [inline stubs/TYPO3/CMS/Frontend/Typolink/EmailLinkBuilder.php:6]: Namespace should be TYPO3\CMS\Frontend\Typolink only, right?

File: src/Rector/v12/v0/typo3/UseEmailLinkBuilderRector.php [hunk 3]
```diff
@@ -0,0 +1,91 @@
...
+use PHPStan\Type\ObjectType;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @changelog https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Deprecation-96500-ContentObjectRenderer-getMailTo.html
+ * @see \Ssch\TYPO3Rector\Tests\Rector\v12\v0\UseEmailLinkBuilderRector\UseEmailLinkBuilderRectorTest
+ */
+final class UseEmailLinkBuilderRector extends AbstractRector
```
simonschaufi, line 19: I don't like the "Use" in the class name and would remove it.

config/v12/typo3-120.php
```diff
@@ -20,4 +20,5 @@
         ]);
     $rectorConfig->rule(\Ssch\TYPO3Rector\Rector\v12\v0\ReplacePreviewUrlMethodRector::class);
     $rectorConfig->rule(RenameExtTypoScriptFilesFileRector::class);
+    $rectorConfig->rule(\Ssch\TYPO3Rector\Rector\v12\v0\UseEmailLinkBuilderRector::class);
```
simonschaufi, line 23: Here is still the old name

helsner: ah this is wrong either way as the \typo3\ is missing in the namespace.. that's some structural issue with the typo3-generate i reported already..

simonschaufi: The wrong namespace should be fixed with latest version

simonschaufi (approved):

PR #3013, opened 2022-06-05
[composer] remove php-cs-fixer, already in ECS, fix newer php-parser for tests

TomasVotruba:
Native php-parser is loaded from phpstan, which is now older version 4.14.
To test php-parser 5.0-dev, we need to preload if from local vendor.

See https://github.com/rectorphp/rector-src/pull/2437

### DIFF
--- a/src/Rector/v9/v5/ExtbaseCommandControllerToSymfonyCommandRector.php
+++ b/src/Rector/v9/v5/ExtbaseCommandControllerToSymfonyCommandRector.php
@@ -7,6 +7,7 @@ namespace Ssch\TYPO3Rector\Rector\v9\v5;
 use Nette\Utils\Strings;
 use PhpParser\Node;
 use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Name;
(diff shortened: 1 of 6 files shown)


src/Rector/v9/v5/ExtbaseCommandControllerToSymfonyCommandRector.php
```diff
@@ -75,6 +76,10 @@ public function refactor(Node $node): ?Node
             return null;
         }
 
+        if (! $node->namespacedName instanceof Name) {
+            return null;
+        }
+
```
TomasVotruba, line 82: PHPStan 4.14 CI fix  :wink:

simonschaufi: Thanx!

TomasVotruba: Yay! You're welcome 🤗

Issue #2955, opened 2022-05-06
Deprecation: #97531 - Context-related methods within TypoScriptFrontendController


sabbelasichon:
Deprecation: #97531 - Context-related methods within TypoScriptFrontendController

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-97531-ContextRelatedMethodsWithinTSFE.html
.. include:: /Includes.rst.txt

=================================================================================
Deprecation: #97531 - Context-related methods within TypoScriptFrontendController
=================================================================================

See :issue:`97531`

Description
===========

One of the main classes within TYPO3 Frontend —
:php:`TypoScriptFrontendController` a.k.a. :php:`$GLOBALS['TSFE']` —
had various short-hand functionality to access the Context API.

This is mainly for historical reasons, before the Context API
was introduced in TYPO3 v9.

For this reason, the related methods have been marked as deprecated:

* :php:`initUserGroups()`
* :php:`isUserOrGroupSet()`
* :php:`isBackendUserLoggedIn()`
* :php:`doWorkspacePreview()`
* :php:`whichWorkspace()`

Impact
======

Calling the methods directly will trigger a PHP deprecation warning.

Affected Installations
======================

TYPO3 installations with custom extensions using one of the methods.
The extension scanner will report any usage as weak match.

Migration
=========

Migrate towards the Context API instead:

.. code-block:: php

    // Is this request within a Workspace currently
    $context->getPropertyFromAspect('workspace', 'isOffline', false);
    // Is a frontend user logged in
    $context->getPropertyFromAspect('frontend.user', 'isLoggedIn', false);

.. index:: Frontend, FullyScanned, ext:frontend

PR #3014, opened 2022-06-05
[FEATURE] Replace TSFE methods with according Context methods

helsner:
Resolves: #2955


config/v12/typo3-120.php
```diff
@@ -25,4 +25,5 @@
     $rectorConfig->rule(\Ssch\TYPO3Rector\Rector\v12\v0\typo3\ReplaceTSFECheckEnableFieldsRector::class);
     $rectorConfig->rule(ReplaceContentObjectRendererGetMailToWithEmailLinkBuilderRector::class);
     $rectorConfig->rule(RemoveUpdateRootlineDataRector::class);
+    $rectorConfig->rule(\Ssch\TYPO3Rector\Rector\v12\v0\typo3\ReplaceTSFEWithContextMethodsRector::class);
```
simonschaufi, line 28: you can import that namespace to be consistent.

simonschaufi (changes requested):

config/v12/typo3-120.php
sabbelasichon: I am gonna merge. Shouldn´t be a manually task. Should be done by rector. Will have a look why it is not changed.

sabbelasichon:
@helsner I have double checked it, looks good to me.

Issue #2916, opened 2022-05-06
Breaking: #97214 - Use UploadedFile objects instead of $_FILES


sabbelasichon:
Breaking: #97214 - Use UploadedFile objects instead of $_FILES

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-97214-UseUploadedFileObjectsInsteadOf_FILES.html
.. include:: /Includes.rst.txt

==============================================================
Breaking: #97214 - Use UploadedFile objects instead of $_FILES
==============================================================

See :issue:`97214`

Description
===========

The TYPO3 request already contains a "disentangled" array of UploadedFile
objects. With this change, these UploadedFile objects are now used instead
of the superglobal :php:`$_FILES` in extbase requests.

Additionally, the FAL ResourceStorage has been adjusted for handling
UploadedFile objects and the ExtensionManager upload handling has been
adjusted.

The next step would be to further adjust FAL to use only PSR provided
methods for handling uploaded files and implementing an API for file
uploads in Extbase.

Impact
======

The global :php:`$_FILES` object is not used in extbase or the extension
manager anymore, instead the PSR request is used.

Affected Installations
======================

All installations extending the TYPO3 Core ResourceStorage object and
overwriting the :php:`addUploadedFile` method.

Migration
=========

Extension authors extending the TYPO3 Core resource storage and implementing
their own handling of :php:`addUploadedFile` need to allow objects of type
:php:`UploadedFile` in addition to the old array from global :php:`$_FILES`.

To do so, switch the type annotation to :php:`array|UploadedFile` and add code that
handles :php:`UploadedFile` objects and arrays.

Example
^^^^^^^

.. code-block:: php

    if ($uploadedFileData instanceof UploadedFile) {
        $localFilePath = $uploadedFileData->getTemporaryFileName();
        if ($targetFileName === null) {
            $targetFileName = $uploadedFileData->getClientFilename();
        }
        $size = $uploadedFileData->getSize();
    } else {
        $localFilePath = $uploadedFileData['tmp_name'];
        if ($targetFileName === null) {
            $targetFileName = $uploadedFileData['name'];
        }
        $size = $uploadedFileData['size'];
    }

.. index:: PHP-API, NotScanned, ext:extbase

PR #3015, opened 2022-06-06
[FEATURE] Add hint to adapt addUploadedFile override

helsner:
Resolves: #2916

### DIFF
--- a/src/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector.php
+++ b/src/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector.php
@@ -1,0 +1,116 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\Rector\v12\v0\typo3;
+
+use PhpParser\Comment;
+use PhpParser\Node;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\UnionType;
+use Rector\Core\Console\Output\RectorOutputStyle;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @changelog https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Breaking-97214-UseUploadedFileObjectsInsteadOf_FILES.html
+ * @see \Ssch\TYPO3Rector\Tests\Rector\v12\v0\typo3\HintNecessaryUploadedFileChangesRector\HintNecessaryUploadedFileChangesRectorTest
+ */
+final class HintNecessaryUploadedFileChangesRector extends AbstractRector
+{
+    /**
+     * @var string
+     */
+    private const MESSAGE = 'When extending the Core ResourceStorage, the addUploadedFile() method needs some adaptions. See Breaking-97214-UseUploadedFileObjectsInsteadOf_FILES.html for full migration advise';
+
+    /**
+     * @var string
+     */
+    private const COMMENT = '// FIXME: Rector: https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Breaking-97214-UseUploadedFileObjectsInsteadOf_FILES.html';
+
+    public function __construct(
+        private readonly RectorOutputStyle $rectorOutputStyle
+    ) {
+    }
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [Class_::class];
+    }
+
+    /**
+     * @param Class_ $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        if ($this->shouldSkip($node)) {
+            return null;
+        }
+
+        $affectedMethod = $node->getMethod('addUploadedFile');
+
+        if (! $affectedMethod instanceof ClassMethod) {
+            return null;
+        }
+
+        $params = $affectedMethod->getParams();
+        if ($params[0]->type instanceof UnionType) {
+            return null;
+        }
+
+        if ($params[0]->type instanceof Identifier && 'array' === $params[0]->type->name) {
+            $comments = $affectedMethod->getComments();
+            $comments = array_filter(
+                $comments,
+                static fn (Comment $comment) => ! str_starts_with($comment->getText(), '// FIXME: Rector:')
+            );
+
+            $comments[] = new Comment(self::COMMENT);
+            $affectedMethod->setAttribute('comments', $comments);
+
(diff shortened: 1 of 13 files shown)


helsner:
So this is basically some kind of prototype..
I thought about this change and came to the conclusion that adapting code would be to difficult, as I see the use cases so specific at this point that we cannot provide a generic and working rewrite.
That's why i wanted to at least add a message.

I'm open for your opinions here.
Especially testing this is tricky as there is no real change

stubs/TYPO3/CMS/Core/Http/UploadedFile.php
```diff
@@ -0,0 +1,11 @@
...
+namespace TYPO3\CMS\Core\Http;
+
+if (class_exists('TYPO3\CMS\Core\Http\UploadedFile')) {
+    return;
+}
+
+class UploadedFile
+{
+
+}
```
simonschaufi, line 11: new line at EOF missing

stubs/TYPO3/CMS/Core/Resource/ResourceStorage.php
```diff
@@ -0,0 +1,11 @@
...
+namespace TYPO3\CMS\Core\Resource;
+
+if (class_exists('TYPO3\CMS\Core\Resource\ResourceFactory')) {
+    return;
+}
+
+class ResourceStorage
+{
+
+}
```
simonschaufi, line 11: new line at EOF missing

tests/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,29 @@
...
+-----
+<?php
+
+namespace Ssch\TYPO3Rector\Tests\Rector\v12\v0\typo3\HintNecessaryUploadedFileChangesRector\Fixture;
+
+use TYPO3\CMS\Core\Resource\ResourceStorage;
+
+final class CustomResourceStorage extends ResourceStorage
+{
+    public function addUploadedFile(array $uploadedFileData)
```
simonschaufi, line 24: where is that comment?

tests/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector/Fixture/fixture_no_apply_needed.php.inc
```diff
@@ -0,0 +1,31 @@
...
+<?php
+
+namespace Ssch\TYPO3Rector\Tests\Rector\v12\v0\typo3\HintNecessaryUploadedFileChangesRector\Fixture;
+
+use TYPO3\CMS\Core\Http\UploadedFile;
+use TYPO3\CMS\Core\Resource\ResourceStorage;
+
+final class UpdatedResourceStorage extends ResourceStorage
+{
+    public function addUploadedFile(array|UploadedFile $uploadedFileData)
```
simonschaufi, line 26: where is that comment?

simonschaufi (changes requested):

tests/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector/Fixture/fixture.php.inc
helsner: I added that comment to enforce the code samples to be valid.
But yes, we could think of the message to be added into the function to hint not only via CLI, but also within the method for the custom changes

stubs/TYPO3/CMS/Core/Http/UploadedFile.php
helsner: i had cases where it just wasn't displayed in github? Maybe that's the issue?

tests/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector/Fixture/fixture.php.inc
simonschaufi: That would be great. I thought this is the case on other rectors as well but they are all the same. A `TODO: ...` comment would be great.

stubs/TYPO3/CMS/Core/Http/UploadedFile.php
simonschaufi: In this case it is really missing.

tests/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector/Fixture/fixture.php.inc
helsner: will try that then.

simonschaufi: I would put the comment at the end of the line if possible and then check if a `TODO: ` comment exists already before adding it.

stubs/TYPO3/CMS/Core/Http/UploadedFile.php
helsner: Yes i got it enabled

tests/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector/Fixture/fixture.php.inc
simonschaufi: Easy, bro!

simonschaufi: Ups, I just noticed that I kinda messed up your last force push. Sorry. I adjusted my code according to yours, just slightly different.

simonschaufi:
The failing test is a false positive. It looks like a bug in rector core, but this `EndBug/add-and-commit` workflow is driving me crazy as it is not working at all. @sabbelasichon

src/Rector/v12/v0/typo3/HintNecessaryUploadedFileChangesRector.php
```diff
@@ -0,0 +1,116 @@
...
+        $params = $affectedMethod->getParams();
+        if ($params[0]->type instanceof UnionType) {
+            return null;
+        }
+
+        if ($params[0]->type instanceof Identifier && 'array' === $params[0]->type->name) {
+            $comments = $affectedMethod->getComments();
+            $comments = array_filter(
+                $comments,
+                static fn (Comment $comment) => ! str_starts_with($comment->getText(), '// FIXME: Rector:')
```
helsner, line 71: with this check i would expect a test fixture which actively checks this.
What do you think?

simonschaufi: done

helsner:
I think it's funny that this should "only be hinting" and now it's bigger than other rules :D 
As PR author i cannot approve, but from my point of view this is fine and can be merged.
Thanks a lot Simon for the rescue!

simonschaufi:
@helsner you can suggest a better name if you come up with something.

simonschaufi (approved):

Issue #2953, opened 2022-05-06
Deprecation: #97354 - ExpressionBuilder methods andX() and orX()


sabbelasichon:
Deprecation: #97354 - ExpressionBuilder methods andX() and orX()

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-97354-ExpressionBuilderMethodsAndXAndOrX.html
.. include:: /Includes.rst.txt

================================================================
Deprecation: #97354 - ExpressionBuilder methods andX() and orX()
================================================================

See :issue:`97354`

Description
===========

`doctrine/dbal` `deprecated`_  the :php:`ExpressionBuilder` methods
:php:`andX()` and :php:`orX`. Therefore, those methods have also been
deprecated in the core facade class (:php:`\TYPO3\CMS\Core\Database\Query\Expression\ExpressionBuilder`),
to avoid shifting too far away.

.. _`deprecated`: https://github.com/doctrine/dbal/commit/84328cd947706210caebcaea3ca0394b3ebc4673

Impact
======

Using :php:`ExpressionBuilder->andX()` and :php:`ExpressionBuilder->orX()`
will raise a deprecation level log entry when called.

Affected Installations
======================

All installations, using the deprecated methods :php:`ExpressionBuilder->andX()`
and :php:`ExpressionBuilder->orX()` in custom extension code. The extension
scanner will detect any usage as weak match.

Migration
=========

Extensions should use the corresponding replacement:

- :php:`ExpressionBuilder->andX()` -> :php:`ExpressionBuilder->and()`
- :php:`ExpressionBuilder->orX()` -> :php:`ExpressionBuilder->or()`

.. note::

   The replacement methods have already been added in a forward-compatible way
   in TYPO3 v11. Thus giving extension developers the ability to adopt new
   methods and still being able to support multiple core versions without
   workarounds.

For example, the following select query:

.. code-block:: php

    $rows = $queryBuilder
      ->select(...)
      ->from(...)
      ->where(
        $queryBuilder->expr()->andX(...),   // replace with and(...)
        $queryBuilder->expr()->orX(...)     // replace with or(...)
      )
      ->executeQuery()
      ->fetchAllAssociative();

should be replaced with:

.. code-block:: php

    $rows = $queryBuilder
      ->select(...)
      ->from(...)
      ->where(
        $queryBuilder->expr()->and(...), // replacement for andX(...)
[...]

PR #3016, opened 2022-06-06
[FEATURE] Substitute ExpressionBuilder methods andX() & orX()

helsner:
Resolves: #2953

### DIFF
--- a/src/Rector/v12/v0/typo3/ReplaceExpressionBuilderMethodsRector.php
+++ b/src/Rector/v12/v0/typo3/ReplaceExpressionBuilderMethodsRector.php
@@ -54,8 +54,8 @@ $rows = $queryBuilder
   ->select(...)
   ->from(...)
   ->where(
-    $queryBuilder->expr()->andX(...),   // replace with and(...)
-    $queryBuilder->expr()->orX(...)     // replace with or(...)
+    $queryBuilder->expr()->andX(...),
+    $queryBuilder->expr()->orX(...)
   )
   ->executeQuery()
   ->fetchAllAssociative();
@@ -66,8 +66,8 @@ $rows = $queryBuilder
   ->select(...)
   ->from(...)
   ->where(
-    $queryBuilder->expr()->and(...), // replacement for andX(...)
-    $queryBuilder->expr()->or(...)   // replacement for orX(...)
+    $queryBuilder->expr()->and(...),
+    $queryBuilder->expr()->or(...)
   )
   ->executeQuery()
   ->fetchAllAssociative();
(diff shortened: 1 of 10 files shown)


src/Rector/v12/v0/typo3/ReplaceExpressionBuilderMethodsRector.php
```diff
@@ -0,0 +1,85 @@
...
+  ->fetchAllAssociative();
+CODE_SAMPLE
+        )]);
+    }
+
+    private function shouldSkip(MethodCall $methodCall): bool
+    {
+        return ! $this->isObjectType(
+            $methodCall,
+            new ObjectType('TYPO3\CMS\Core\Database\Query\Expression\ExpressionBuilder')
```
helsner, line 82: In my mind this is the correct ObjectType, but the test says otherwise..
I don't know how to handle this/ how the chaining of methods is working

stubs/TYPO3/CMS/Core/Database/Query/QueryBuilder.php
```diff
@@ -0,0 +1,54 @@
...
+    {
+        return '';
+    }
+
+    public function executeQuery(): QueryBuilder
+    {
+        return $this;
+    }
+
+    public function fetchAllAssociative(): QueryBuilder
```
helsner, line 50: Same as with executeQuery - basically the method itself is wrong in here

stubs/TYPO3/CMS/Core/Database/Query/QueryBuilder.php
```diff
@@ -0,0 +1,54 @@
+<?php
+declare(strict_types=1);
```
simonschaufi, line 2: According to code style, there should be an empty line before and after declare statements.

stubs/TYPO3/CMS/Core/Database/Query/Expression/ExpressionBuilder.php
```diff
@@ -0,0 +1,31 @@
+<?php
+declare(strict_types=1);
```
simonschaufi, line 2: According to code style, there should be an empty line before and after declare statements.

simonschaufi (changes requested):

stubs/TYPO3/CMS/Core/Database/Query/QueryBuilder.php
helsner: It's pretty annoying that some of that stuff seems to not get applied, isn't it?
EOL, FQN in config files and now this

src/Rector/v12/v0/typo3/ReplaceExpressionBuilderMethodsRector.php
helsner: I could adapt one test case, where the ExpressionBuilder is set specifically.
I stepped into the chained methods as well but could not really figure out what's happening there.
On comparison with other rules the chaining is handled by checking for the initial ObjectType, e.g. here: \Ssch\TYPO3Rector\Rector\v11\v0\RemoveLanguageModeMethodsFromTypo3QuerySettingsRector

But seems to be different here.

helsner: So i changed the stubs and now it works?
I'm a little irritated tbh, but the tests are obviously green and the shouldskip method really only checks for the ExpressionBuilder, so i would guess this is a proper and fully working rule without side effects.
Just wondering what was not fitting before that..

Issue #2949, opened 2022-05-06
Deprecation: #97244 - CompositeExpression methods 'add()' and 'addMultiple()'


sabbelasichon:
Deprecation: #97244 - CompositeExpression methods 'add()' and 'addMultiple()'

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-97244-CompositeExpressionMethodsAddAndAddMultiple.html
.. include:: /Includes.rst.txt

=============================================================================
Deprecation: #97244 - CompositeExpression methods 'add()' and 'addMultiple()'
=============================================================================

See :issue:`97244`

Description
===========

`doctrine/dbal` `deprecated`_ multiple :php:`CompositeExpression` methods
:php:`CompositeExpression->add()` and :php:`CompositeExpression->addMultiple()`.
Therefore, those methods have also been deprecated in the core facade class
(:php:`\TYPO3\CMS\Core\Database\Query\Expression\CompositeExpression`),
to avoid shifting too far away.

.. _`deprecated`: https://github.com/doctrine/dbal/commit/7bcd6ebcc2d30ba96cf00d3dca2345d6ae779cf9

Impact
======

Using :php:`CompositeExpression->add()` and :php:`CompositeExpression->addMultiple()`
will raise a deprecation level log entry when called.

Affected Installations
======================

In general, instances with extensions that use the deprecated methods
:php:`CompositeExpression->add()` and :php:`CompositeExpression->addMultiple()`.
The extension scanner finds usages of :php:`CompositeExpression->addMultiple()`
with a weak match. The other method is not scanned, as its name is too common.

Migration
=========

The deprecated methods :php:`CompositeExpression->add()` and :php:`CompositeExpression->addMultiple()`
can be replaced by using the new method :php:`CompositeExpression->with()`.

.. note::

    The replacement method :php:`CompositeExpression->with()` has already been
    added in a forward-compatible way in TYPO3 v11. Thus giving extension developers
    the ability to adopt new methods and still being able to support multiple core
    versions without workarounds.

For example, the following code:

.. code-block:: php

    use TYPO3\CMS\Core\Database\Query\Expression\CompositeExpression;

    $compositeExpression = CompositeExpression::or();

    $compositeExpression->add(
      $queryBuilder->expr()->eq(
        'field',
        $queryBuilder->createNamedParameter($singleValue)
      )
    );
    $compositeExpression->addMultiple(
      [
[...]

PR #3021, opened 2022-06-10
[FEATURE] Replace add() and addMultiple() of CompositeExpression with with()

helsner:
Resolves: #2949


simonschaufi (approved):
Well done!

Issue #3023, opened 2022-06-10
ForwardResponseInsteadOfForwardMethodRector does not validate return type


helsner:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.13.0             |

Rule does not adapt return type causing fatal error on second rund

## Minimal PHP Code Causing Issue
```
public function listAction(): void
   {
        $this->forward('show');
   }
```

## Expected Behaviour
```
public function listAction(): ResponseInterface
   {
        return new ForwardResponse('show');
   }
```

what actually happens:
```
public function listAction(): void
   {
        return new ForwardResponse('show');
   }
```

helsner:
i provided a testcase - it seems to be related if there are params existing or not.

PR #3024, opened 2022-06-10
[BUGFIX] Resolve ReturnType 'void' along with DocComment

helsner:
Relates: #3023

### DIFF
--- a/src/Rector/v11/v0/ForwardResponseInsteadOfForwardMethodRector.php
+++ b/src/Rector/v11/v0/ForwardResponseInsteadOfForwardMethodRector.php
@@ -4,9 +4,11 @@ declare(strict_types=1);
 
 namespace Ssch\TYPO3Rector\Rector\v11\v0;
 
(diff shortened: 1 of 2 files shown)


simonschaufi (approved):

simonschaufi:
@helsner I'm fine with the change itself but I'm sad that phpstan is broken again :cry: 

I would prefer to first fix the phpstan issue, then rebase this patch to have everything green again.

helsner:
Yeah, I'm currently on another rule with the same issue after composer update
Seems like something got updated and now the whole chain breaks
I'm already looking but did not find anything yet

helsner:
it breaks with "symplify/phpstan-rules": "10.3.1",
10.3.0 works so yeah

simonschaufi:
I'm therefore happy with making the version sticky because dependencies could easily crash our pipeline anytime, which is really annoying.

helsner:
I will do that then.

Issue #2921, opened 2022-05-06
Breaking: #97312 - Remove context sensitive help


sabbelasichon:
Breaking: #97312 - Remove context sensitive help

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-97312-RemoveContextSensitiveHelp.html
.. include:: /Includes.rst.txt

================================================
Breaking: #97312 - Remove context sensitive help
================================================

See :issue:`97312`

Description
===========

The arguments for removing context sensitive help were:

* The help was not really context sensitive, it only relied on tablename
  and fieldname, if a field was used for different purposes in different
  content types, the CSH always showed the same help
* There was outdated information in CSH (e.g. Screenshots form TYPO3 4.x)
  and nobody is available to update the information
* Some CSH descriptions explained the same content with different words
  which is confusing (e.g. tt_content - CType > Title: "Type" > CSH
  Tooltip: "Select the kind of Page Content this element represents.
  New options will appear when you save the record.")
* Many CSH texts did not provide useful additional information
  (e.g. tt_content - header > Title: "Header" > CSH Tooltip:
  "Enter header text for the Content Element.")
* The available online documentation https://docs.typo3.org/ improved
  a lot and helps better than the CSH in most cases, as it is up to date
* CSH was hidden for most users, as it was only available by clicking on
  a label (no hint that help was available without hovering the label by
  mouse, not available for keyboard users)
* `description` is available for explanations when they are required.
  Adding relevant information as `description` will help everyone as it is
  visible.
* The removal was already proposed in 2019 (see
  https://decisions.typo3.org/t/drop-context-sensitive-help-in-core/511)
  and most arguments against removal can be solved using the `description`
  or by linking to the official documentation
* Removal of CSH also removed a lot of outdated files (and results in
  smaller footprint of the TYPO3 core package)

The route `help_cshmanual_popup` has been removed.

Help buttons :php:`Components\Buttons\Action\HelpButton` only return an
empty string and trigger a deprecation warning.

The CSH descriptions are not loaded any longer for tables.

All labels are adjusted to not contain <abbr> tags inside any longer.

The method :php:`cshItem()` of
:php:`TYPO3\CMS\Backend\Utility\BackendUtility` always returns an empty
string and triggers an deprecation warning.

The TYPO3 Manual menu item has been removed and a link to the
TYPO3 Online Documentation has been added to the menu.

The backend display related TCA option
[...]

PR #3025, opened 2022-06-10
[FEATURE] Remove redundant ['interface']['always_description']

helsner:
Resolves: #2921


config/v12/tca-120.php
```diff
@@ -10,4 +10,5 @@
     $rectorConfig->import(__DIR__ . '/../config.php');
     $rectorConfig->rule(MigrateNullFlagRector::class);
     $rectorConfig->rule(MigrateRequiredFlagRector::class);
+    $rectorConfig->rule(\Ssch\TYPO3Rector\Rector\v12\v0\tca\RemoveTCAInterfaceAlwaysDescriptionRector::class);
```
simonschaufi, line 13: please try to remember to import that namespace manually.

helsner: ah yes

Issue #2870, opened 2022-05-06
Breaking: #96222 - Add getOptions() to WidgetInterface


sabbelasichon:
Breaking: #96222 - Add getOptions() to WidgetInterface

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96222-AddGetOptionsToWidgetInterface.html
.. include:: /Includes.rst.txt

======================================================
Breaking: #96222 - Add getOptions() to WidgetInterface
======================================================

See :issue:`96222`

Description
===========

With :issue:`93210` the dashboard was extended for the functionality to
refresh single widgets. This also required to extend the :php:`WidgetInterface`.
To stick to TYPO3's backwards compatibility promise, the new method was
commented out and instead a :php:`methodExists()` check performed.

This now has changed. The check is removed and the :php:`WidgetInterface`
now forces the presence of the :php:`getOptions()` method in all widgets.

Impact
======

All dashboard widgets are now forced to implement the :php:`getOptions()`
method, returning the widget options. Otherwise this will cause a PHP
fatal error.

Affected Installations
======================

All installations using custom dashboard widgets.

Migration
=========

Add the :php:`getOptions()` method to all of your custom widget classes.

.. code-block:: php

    public function getOptions(): array
    {
        return $this->options;
    }

.. index:: Backend, PHP-API, NotScanned, ext:dashboard

PR #3027, opened 2022-06-12
[FEATURE] Add method according to WidgetInterface integration

helsner:
Resolves: #2870

### DIFF
--- a/src/Rector/v12/v0/typo3/AddMethodToWidgetInterfaceClassesRector.php
+++ b/src/Rector/v12/v0/typo3/AddMethodToWidgetInterfaceClassesRector.php
@@ -1,0 +1,138 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\Rector\v12\v0\typo3;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Identifier;
+use PhpParser\Node\Param;
+use PhpParser\Node\Stmt\Class_;
+use PhpParser\Node\Stmt\ClassMethod;
+use PhpParser\Node\Stmt\Return_;
+use PHPStan\Reflection\ReflectionProvider;
+use PHPStan\Type\ObjectType;
+use Rector\Core\Rector\AbstractRector;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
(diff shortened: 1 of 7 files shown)


helsner:
This needs feedback on the TODO marker i set.

stubs/TYPO3/CMS/Dashboard/Widgets/WidgetInterface.php
```diff
@@ -0,0 +1,15 @@
...
+if (interface_exists('TYPO3\CMS\Dashboard\Widgets\WidgetInterface')) {
+    return;
+}
+
+interface WidgetInterface
+{
+    public function renderWidgetContent(): string;
+
+    public function getOptions(): array;
+}
```
simonschaufi, line 15: missing EOF

tests/Rector/v12/v0/typo3/AddMethodToWidgetInterfaceClassesRector/Fixture/fixture_with_existing_method.php.inc
```diff
@@ -0,0 +1,22 @@
...
+        return 'foo';
+    }
+
+    public function getOptions(): array
+    {
+        return $this->options;
+    }
+}
+
+?>
```
simonschaufi, line 22: missing EOF

src/Rector/v12/v0/typo3/AddMethodToWidgetInterfaceClassesRector.php
```diff
@@ -0,0 +1,138 @@
...
+        $className = $this->getName($class);
+        if (null === $className) {
+            return true;
+        }
+
+        if (! $this->reflectionProvider->hasClass($className)) {
+            return true;
+        }
+
+        // TODO: as far as i understand this should be active, but does currently prevent the code migration
```
simonschaufi, line 119: mhh, I can't say anything about that.

simonschaufi (changes requested):

src/Rector/v12/v0/typo3/AddMethodToWidgetInterfaceClassesRector.php
helsner: then i guess we need some expert advice from @sabbelasichon

sabbelasichon: The problem is that the code is technically broken, because the contract is not fullfilled at runtime. We should not use the reflectionProvider but uses somehow static reflection with PhpParser nodes.

sabbelasichon:
@helsner That´s basically the change i did: https://github.com/sabbelasichon/typo3-rector/pull/3027/files#diff-bccc36579ba79f20a12679c7d6d64eab62005e4a7431748ddde2c99de56f1f27R98

PR #3030, opened 2022-06-15
[FEATURE] Provide components handling the fe_login_mode

helsner:
Resolves: #2885


helsner:
This is basically a copy of the previous #3028 which is totally destroyed with my force push etc.

Issue #3039, opened 2022-06-27
SubstituteBackendTemplateViewWithModuleTemplateRector does not ignore initializeAction


simonschaufi:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version | 0.13.0     |

## Minimal PHP Code Causing Issue

Original:

```php
    public function initializeAction()
    {
        // some code
    }
```

Actual output:

```php
    public function initializeAction(): \Psr\Http\Message\ResponseInterface
    {
        $moduleTemplate = $this->moduleTemplateFactory->create($this->request);
        // some code
        $moduleTemplate->setContent($this->view->render());
        return $this->htmlResponse($moduleTemplate->renderContent());
    }
```

## Expected Behaviour

initializeAction is a special action which doesn't need to return a response but just initializes an actual action. Same goes for 
"initializeIndexAction". Don't touch these methods!

helsner:
I'm a little bit irritated.
Shouldn't that be task of the `ForwardResponseInsteadOfForwardMethodRector` ?

simonschaufi:
Well, I ran every v11 rule individually and this rule messed it up.

helsner:
Yep, can confirm

simonschaufi:
So basically there should be a check if the method starts with "initialize" and skip.

helsner:
will add yeah

PR #3045, opened 2022-06-28
[BUGFIX] Prevent wrongful change of initialize methods

helsner:
Resolves: #3039


src/Rector/v11/v5/SubstituteBackendTemplateViewWithModuleTemplateRector.php
```diff
@@ -219,7 +219,9 @@ private function callSetContentAndGetContent(ClassMethod $classMethod): void
     {
         $classMethodName = (string) $this->getName($classMethod->name);
 
-        if (! str_ends_with($classMethodName, 'Action')) {
+        if (! str_ends_with($classMethodName, 'Action')
+            || 'initializeAction' === $classMethodName
+            || 'initializeIndexAction' === $classMethodName) {
```
simonschaufi, line 224: no, this won't fix it. The pattern is "initialize" + whatever + "Action". So we just need to test if the action starts with "initialize" and ends with "Action".

simonschaufi (changes requested):
still broken

simonschaufi (approved):

PR #3047, opened 2022-06-29
TASK: Use rector/rector instead of rector/rector-src

sabbelasichon:


config/v8/typo3-83.php
```diff
@@ -13,7 +12,6 @@
 
     $rectorConfig->rule(RefactorMethodFileContentRector::class);
     $rectorConfig->rule(RefactorQueryViewTableWrapRector::class);
-    $rectorConfig->rule(IconsRector::class);
```
helsner, line 16: why removed? 🤔

rector.php
```diff
@@ -20,7 +20,7 @@
 
     $rectorConfig->parallel();
     $rectorConfig->importNames();
-    $rectorConfig->disableImportShortClasses();
+    $rectorConfig->importShortClasses(false);
```
simonschaufi, line 23: bitte hier auch das template dafür anpassen!

.gitattributes
```diff
@@ -24,3 +24,4 @@ utils/generator export-ignore
 utils/rules export-ignore
 typo3.constants.php export-ignore
 /stubs export-ignore
+/bin export-ignore
```
simonschaufi, line 27: are you sure this needs to be excluded?

sabbelasichon: Yes, it is only for the generate command. I will show you. We had to create our custom console application.

simonschaufi: ah ok so only relevant when using the git source.

Issue #3071, opened 2022-07-22
Remove rule QueryLogicalOrAndLogicalAndToArrayParameterRector


froemken:
# Bug Report

| Subject        | Details                                                         |
| :------------- | :---------------------------------------------------------------|
| Rector version |  0.12.23             |

**Issue**

Please remove rule `QueryLogicalOrAndLogicalAndToArrayParameterRector`

Rector want's me to update my code as follows:

```
     {
         $query = $this->createQuery();
         $query->matching(
-            $query->logicalOr(
-                $query->like('name', '%' . trim($name) . '%'),
-                $query->equals('warn_cell_id', $name)
-            )
+            $query->logicalOr([$query->like('name', '%' . trim($name) . '%'), $query->equals('warn_cell_id', $name)])
         );
         return $query->execute()->toArray();
     }
```

Applied rules:
`QueryLogicalOrAndLogicalAndToArrayParameterRector (https://docs.typo3.org/m/typo3/book-extbasefluid/master/en-us/6-Persistence/3-implement-individual-database-queries.html)`

The given link does not contain any `logicalAnd` or `logicalOr` containing an array.

Further you will find following comment in `logicalAnd` and `logicalOr` in Extbase source code:

```
/*
 * todo: Deprecate accepting an array as $constraint1
 *       Add param type hints for $constraint1 and $constraint2
 *       Make $constraint2 mandatory
 *       Add AndInterface return type hint
 *       Adjust method signature in interface
 */
```

So, please removve rule `QueryLogicalOrAndLogicalAndToArrayParameterRector` from rector

Stefan

helsner:
Relates: #2462

helsner:
closed due to reasoning found in PR #3072 
Thanks for the contribution!

simonschaufi:
One can always exclude single rules with:

```php
$rectorConfig->services()->remove(QueryLogicalOrAndLogicalAndToArrayParameterRector::class);
```

PR #3072, opened 2022-07-22
Remove rule QueryLogicalOrAndLogicalAndToArrayParameterRector #3071

froemken:
Resolves: #3071


simonschaufi:
What is your intention, why was this rule removed?

helsner:
Resolves: #3071

froemken:
@simonschaufi Sorry, I have forgotten to add the Issue #3071 in my commit message. I have described it there on detail. Thank you @helsner to put this PR into relation

simonschaufi (approved):

helsner (changes requested):
I would decline this change.
It might seem not logical to you now, but when upgrading the core to v9 this is a necessary change.
Within v10 the method changed into multiple arguments, but in v9 LTS it is not, thus making this rule fully correct in its place.
As there is no change until now it is not wrong to have this active, even though it might irritate one.
In case the TODO segment within the logicalAnd will be integrated there will be a new changelog entry which then will turn into a rule of us.
But for the sake of the whole chain to work this is fitting and necessary

simonschaufi:
@helsner thanks for your detailed code review

simonschaufi:
Well but if it's not an array, all passed arguments are used: https://github.com/TYPO3/typo3/blob/9.5/typo3/sysext/extbase/Classes/Persistence/Generic/Query.php#L417 but still, the IDE will complain about wrong number of arguments, so passing an array is "more correct".

helsner:
Yeah we could do that or exclude the rule within the v9 config.
But in general i would guess this does not break anything and in case this method will change in the future we will have a new rule to rewrite or we can drop it then

simonschaufi:
One can always exclude single rules with:

```php
$rectorConfig->services()->remove(QueryLogicalOrAndLogicalAndToArrayParameterRector::class);
```

sabbelasichon:
The rule confuses at least some people. So does other rules. I would vote for dropping it. Let's focus on the important ones

froemken:
Why you have close this PR? Because the IDE complains about wrong type?!
"this does not break anything" -> that's correct, but it touches code which does not have to be updated. I still prefer to remove that rule.

helsner:
In general it does not matter if the rule is applied or not for a proper update to a current TYPO3 release.
It does though for specific updates to v9.
So the decision is if we see this as use case as the version is relatively old.

For me it would also be ok to drop the rule with that perspective.

simonschaufi:
@helsner Does it really matter for v9? The code in v8.7 looks exactly the same: https://github.com/TYPO3/typo3/blob/TYPO3_8-7/typo3/sysext/extbase/Classes/Persistence/Generic/Query.php#L404

helsner:
we should take the original issue into perspective #2462
The issue seems to be a combination with other tools that want to cleanup the superfluous arguments.
So the reason for the existence if the rule is to prevent harm by deleting the arguments wrongfully/before the actual migration to a newer TYPO3 version

simonschaufi:
Ok, the problem is if you additionally use https://github.com/rectorphp/rector-src/blob/main/rules/Php71/Rector/FuncCall/RemoveExtraParametersRector.php.

I still vote for: keep our rule and let the user exclude it.

simonschaufi (changes requested):
let's keep this rule

helsner:
regarding the reason for the existence of the rule i figured out. i support to keep it.
I prefer to prevent issues caused by rector

helsner:
lolz - for documentation:
https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.0/Breaking-96044-HardenMethodSignatureOfLogicalAndAndLogicalOr.html

Issue #3091, opened 2022-08-09
Breaking: #97752 - MailerAdapterInterface removed


sabbelasichon:
Breaking: #97752 - MailerAdapterInterface removed

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-97752-MailerAdapterInterfaceRemoved.html
.. include:: /Includes.rst.txt

.. _breaking-97752-1654761506:

=================================================
Breaking: #97752 - MailerAdapterInterface removed
=================================================

See :issue:`97752`

Description
===========

The :php:`\TYPO3\CMS\Core\Mail\MailerAdapterInterface` has been removed,
since the interface became unused in v7 due to removal of core's
:php:`SwiftMailerAdapter` implementation, which had been used as hook
subscriber in the also removed :php:`MailUtility::mail()` method.

Impact
======

Implementing the interface in custom extension code will trigger
a PHP Error.

Affected installations
======================

All installations implementing the interface in custom extension code,
which is very unlikely. The extenion scanner will report any usage as
strong match.

Migration
=========

Remove any usage of the interface in extension code.

.. index:: PHP-API, FullyScanned, ext:core

tomasnorre:
I'll give this a spin.

PR #3123, opened 2022-08-26
[TASK] Add rule to remove MailerAdapterInterface

tomasnorre:
Resolves: #3091


tomasnorre:
It's been quite a while since my last contribution, so please provide some feedback.
Thanks.

sabbelasichon:
@tomasnorre Very good. I think the only thing missing is adding this rule to the config

tomasnorre:
Sorry, forgot about that. It's done now.

Issue #3105, opened 2022-08-09
Deprecation: #97787 - Severities of flash messages and reports deprecated


sabbelasichon:
Deprecation: #97787 - Severities of flash messages and reports deprecated

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-97787-SeveritiesOfFlashMessagesAndReportsDeprecated.html
.. include:: /Includes.rst.txt

.. _deprecation-97787-1655495192:

=========================================================================
Deprecation: #97787 - Severities of flash messages and reports deprecated
=========================================================================

See :issue:`97787`

Description
===========

With the introduction of :php:`\TYPO3\CMS\Core\Type\ContextualFeedbackSeverity`,
the existing severity constants of :php:`\TYPO3\CMS\Core\Messaging\FlashMessage`
and :php:`\TYPO3\CMS\Reports\Status` have been marked as deprecated.

Impact
======

Passing the constants as listed below to the constructor of
:php:`\TYPO3\CMS\Core\Messaging\FlashMessage` will trigger a deprecation level
log entry:

* :php:`\TYPO3\CMS\Core\Messaging\FlashMessage::NOTICE`
* :php:`\TYPO3\CMS\Core\Messaging\FlashMessage::INFO`
* :php:`\TYPO3\CMS\Core\Messaging\FlashMessage::OK`
* :php:`\TYPO3\CMS\Core\Messaging\FlashMessage::WARNING`
* :php:`\TYPO3\CMS\Core\Messaging\FlashMessage::ERROR`

Passing the constants as listed below to the constructor of
:php:`\TYPO3\CMS\Reports\Status` will trigger a deprecation level log entry:

* :php:`\TYPO3\CMS\Reports\Status::NOTICE`
* :php:`\TYPO3\CMS\Reports\Status::INFO`
* :php:`\TYPO3\CMS\Reports\Status::OK`
* :php:`\TYPO3\CMS\Reports\Status::WARNING`
* :php:`\TYPO3\CMS\Reports\Status::ERROR`

Affected installations
======================

All installations with 3rd party plugins using the aforementioned constants are
affected.

Migration
=========

Use the cases of the :php:`\TYPO3\CMS\Core\Type\ContextualFeedbackSeverity` enum.
The following cases are available:

* :php:`\TYPO3\CMS\Core\Type\ContextualFeedbackSeverity::NOTICE`
[...]

PR #3124, opened 2022-08-26
[TASK] Use value property of getSeverity()

### LINKED ISSUE #3092: Breaking: #97787 - AbstractMessage->getSeverity() returns ContextualFeedbackSeverity
Breaking: #97787 - AbstractMessage->getSeverity() returns ContextualFeedbackSeverity

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-97787-AbstractMessageGetSeverityReturnsContextualFeedbackSeverity.html
.. include:: /Includes.rst.txt

.. _breaking-97787-1657629392:

====================================================================================
Breaking: #97787 - AbstractMessage->getSeverity() returns ContextualFeedbackSeverity
====================================================================================

See :issue:`97787`

Description
===========

The class :php:`\TYPO3\CMS\Core\Messaging\AbstractMessage` and the extended
class :php:`\TYPO3\CMS\Core\Messaging\FlashMessage` both have a method
:php:`getSeverity()` to return a flash message's severity. The return type of
the method is changed to return an instance of :php:`\TYPO3\CMS\Core\Type\ContextualFeedbackSeverity`.

As this method isn't supposed to be used publicly, it is declared `internal` now.

Impact
======

Relying on the return type of :php:`\TYPO3\CMS\Core\Messaging\AbstractMessage->getSeverity()`
being `int` will throw a :php:`TypeError` exception.

There is no negative impact in the following cases:
* Using the severity enum in Fluid for direct rendering
* Using the severity enum in :php:`json_encode()`
In these cases, the enum's value is automatically used.

Affected installations
======================

All extensions using :php:`\TYPO3\CMS\Core\Messaging\AbstractMessage->getSeverity()`
in PHP are affected, if the integer type is expected.

Migration
=========

If the integer type of :php:`\TYPO3\CMS\Core\Messaging\AbstractMessage->getSeverity()`
is expected, use the :php:`value` property of the :php:`ContextualFeedbackSeverity` enum:

.. code-block:: php

    $flashMessage = new \TYPO3\CMS\Core\Messaging\FlashMessage('This is a message');
    $severityAsInt = $flashMessage->getSeverity()->value;

The same applies to Fluid template, where the severity is used within another
structure, e.g. as an array key:

.. code-block:: html

    <div class="x" class="{severityClassMapping.{status.severity.value}}">
        
    </div>

.. index:: PHP-API, NotScanned, ext:core

tomasnorre:
Resolves #3092


src/Rector/v12/v0/typo3/AbstractMessageGetSeverityRector.php
```diff
@@ -0,0 +1,77 @@
...
+ */
+final class AbstractMessageGetSeverityRector extends AbstractRector
+{
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [MethodCall::class, PropertyFetch::class];
```
helsner, line 27: PropertyFetch can be removed, as we want to detect the MethodCall

src/Rector/v12/v0/typo3/AbstractMessageGetSeverityRector.php
```diff
@@ -0,0 +1,77 @@
...
+    }
+
+    /**
+     * @param MethodCall|PropertyFetch $node
```
helsner, line 31: here as well, this will also resolve the phpstan issues for the shouldSkip method

src/Rector/v12/v0/typo3/AbstractMessageGetSeverityRector.php
```diff
@@ -0,0 +1,77 @@
...
+$severityAsInt = $flashMessage->getSeverity()->value;
+CODE_SAMPLE
+        )]);
+    }
+
+    private function shouldSkip(MethodCall $methodCall): bool
+    {
+        if (! $this->nodeTypeResolver->isMethodStaticCallOrClassMethodObjectType(
+            $methodCall,
+            new ObjectType('TYPO3\CMS\Core\Messaging\FlashMessage')
```
helsner, line 70: FlashMessage is not the only class AFAIK
We had a similar issue already with #3105
They missed out `TYPO3\CMS\Core\Messaging\AbstractMessage` in the docs.
We should add a testcase and rework for that as well, in case it is valid! I'm not 100% sure there

src/Rector/v12/v0/typo3/AbstractMessageGetSeverityRector.php
```diff
@@ -0,0 +1,77 @@
...
+    {
+        if ($this->shouldSkip($node)) {
+            return null;
+        }
+
+        if (! $this->isNames($node->name, ['getSeverity'])) {
+            return null;
+        }
+
+        return null;
```
helsner, line 43: this last segment should then actually apply the change
i would expect something like
`return $this->nodeFactory->createPropertyFetch($node, 'value')`
or maybe `$node->var`, not sure

src/Rector/v12/v0/typo3/AbstractMessageGetSeverityRector.php
tomasnorre: I would need to check that, the rest is working for now. But this could be an valid point.

helsner: Yes, that should be fitting.

tomasnorre:
The tests are failing in files that I haven't touched, due to incomplete tests.

PHPStan is failing due to the latest update:  
- Upgrading phpstan/phpstan (1.8.2 => 1.8.5)

This is the PHPStan update locally. But I don't know if it's from 1.8.2 to 1.8.3 or where the "problem" lies. With 1.8.2 PHPstan works locally. 

How should I deal with this?

sabbelasichon:
@tomasnorre Thanks Tomas. I have updated PHPStan and Rector and the errors are gone.

Issue #3134, opened 2022-09-21
Readd segment to prevent FQN for older core versions


helsner:
As upgrades often go step by step we should readd a commented segment.
This should prevent the FQN change in the files:

```
'ext_localconf.php',
'ext_tables.php',
'ClassAliasMap.php',
__DIR__ . '/**/Configuration/*.php',
__DIR__ . '/**/Configuration/**/*.php',
```

helsner:
i found the commit, but don't quite understand the fixture to negate the FQN config..
https://github.com/sabbelasichon/typo3-rector/commit/6c84f1566ef28b04c0735501c72cdd628553e00a#diff-1e5b8e5b50c697439d95db3b37311c2bbf5468460aa3714551195c2f5dd1c94b

Patta:
i can confirm this and just skip the config files in https://github.com/sabbelasichon/typo3-rector/blob/292d7cda3304dff9839a96012e921d6e6334dcfb/templates/rector.php.dist#L46

sabbelasichon:
We should add those to the skip section. The PostRectors are not easily possible anymore, because it is not part of the public api of rector. 
I will create a PR for this.

PR #3141, opened 2022-09-27
TASK: Add skip configuration for NameImportingPostRector

sabbelasichon:
Resolves: #3134


helsner (approved):
Adding that in the config docs segment would be nice as well

Issue #3417, opened 2023-05-26
BUG: Uncaught TypeError: Ssch\TYPO3Rector\Helper\OldSeverityToLogLevelMapper::mapSeverityToLogLevel


spoonerWeb:
## Problem

Doing an upgrade of project extensions from v7 to v11, I got this error:
```
PHP Fatal error: Uncaught TypeError: Ssch\TYPO3Rector\Helper\OldSeverityToLogLevelMapper::mapSeverityToLogLevel()
```
Issue is a method with `int` needed, but `string` given. 
src/Rector/v9/v0/SubstituteGeneralUtilityDevLogRector.php:70

sabbelasichon:
Yes, this could be the case in the ancient times. I have fixed it.

PR #3418, opened 2023-05-26
Update SubstituteGeneralUtilityDevLogRector.php

spoonerWeb:
Parse $severity to int as it comes as string in older versions < v9.

Closes: #3417


sabbelasichon:
@spoonerWeb I am gonna close this one in favour of #3429. You have to cast the final value to an int.

PR #3424, opened 2023-06-06
[FEATURE] Provide rules to migrate EXT:vhs viewhelpers

kitzberger:
* v:extension.path.relative -> f:uri.resource
* v:format.json.encode -> f:format.json
* v:media.image -> f:media.image
* v:uri.image -> f:uri.image
* v:l -> f:translate
* v:variable.set -> f:variable
* v:or -> inline ternary operator

t.b.c.

### DIFF
--- a/config/replace-vhs.php
+++ b/config/replace-vhs.php
@@ -3,6 +3,7 @@
 declare(strict_types=1);
 
 use Rector\Config\RectorConfig;
+use Ssch\TYPO3Rector\FileProcessor\Fluid\Rector\vhs\ReplaceExtensionPathRelativeFluidRector;
 use Ssch\TYPO3Rector\FileProcessor\Fluid\Rector\vhs\ReplaceFormatJsonEncodeFluidRector;
 use Ssch\TYPO3Rector\FileProcessor\Fluid\Rector\vhs\ReplaceLFluidRector;
 use Ssch\TYPO3Rector\FileProcessor\Fluid\Rector\vhs\ReplaceMediaImageFluidRector;
@@ -12,6 +13,7 @@ use Ssch\TYPO3Rector\FileProcessor\Fluid\Rector\vhs\ReplaceVariableSetFluidRecto
 
 return static function (RectorConfig $rectorConfig): void {
     $rectorConfig->import(__DIR__ . '/config.php');
+    $rectorConfig->rule(ReplaceExtensionPathRelativeFluidRector::class);
     $rectorConfig->rule(ReplaceFormatJsonEncodeFluidRector::class);
     $rectorConfig->rule(ReplaceLFluidRector::class);
     $rectorConfig->rule(ReplaceMediaImageFluidRector::class);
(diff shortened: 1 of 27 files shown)


helsner:
Nice draft, a note though:
The `v:variable.set -> f:variable` change could cause issues, when used with multi dimensional
so `v:variable.set name="foo.bar"` won't work with `f:variable name="foo.bar"` from my experience

sabbelasichon:
@kitzberger Nice. What we will also need are tests. Could you provide them?

kitzberger:
> @kitzberger Nice. What we will also need are tests. Could you provide them?

@sabbelasichon, like this? Any suggestions? I know, one is failing. I need to figure out how to migrate single viewhelper properties.

kitzberger:
> Nice draft, a note though: The `v:variable.set -> f:variable` change could cause issues, when used with multi dimensional so `v:variable.set name="foo.bar"` won't work with `f:variable name="foo.bar"` from my experience

@helsner, good point. Is there a way of migrating the code and print some warning in such a case? Or is that counter-productive since nobody every reads them anyways?

helsner:
I would say there is no proper way of migration as the logic behind the car cannot be further checked (especially in HTML).
A warning should be possible and the prevention of changing the view helper call in case there is a dot within the name
As we don't change the code then, it should already be good, in case the warning is ignored we don't destroy functionality and I cannot help people that ignore our notice 🤷

kitzberger:
Do you know of any examples where a warning is being output? Thanks for your input :-)

helsner:
> Do you know of any examples where a warning is being output? Thanks for your input :-)

simply look for `rectorOutputStyle` in existing rules.
It's basically used via constructor and `...->warning()` call most of the times

kitzberger:
@helsner, thanks for the hint!

I've now added a more strict pattern to detect the simple cases like: `v:variable.set name="xx"` and migrate only those automatically. Any left occurrences of v:variable.set I keep untouched and print a warning instead.

helsner:
@kitzberger nice, but i would point out to why the change is not made as well, as this might not be known (i found that out the hard way 😂 )

helsner:
This still looks great, but i just checked and noticed that we have some TODO's left

mehdichaouch:
> This still looks great, but i just checked and noticed that we have some TODO's left

Nice PR, do you know when it'll be ready to merge?

Thanks,

sabbelasichon:
@helsner What kind of TODO´s are left here?

helsner:
> @helsner What kind of TODO´s are left here?

Those marked in the code I guess
Maybe we can still merge and they will be added afterwards
But usually we migrate fully because it would cause too much frustration not handling attributes etc

sabbelasichon:
@helsner If you like you can merge it. I am not using VHS anyways ;-)

helsner:
@kitzberger what about you phillipp?
are you ready to merge?

PR #3438, opened 2023-07-04
TASK: Run rector and fix code style

simonschaufi:


simonschaufi:
@helsner This is why all other rector builds are currently failing (and I'm wondering why `EndBug/add-and-commit` is not doing it's job as it should... :shrug:

Issue #3450, opened 2023-08-05
double required added


helsner:
In my talk we noticed that my example demo rector project had a case where a double required has been added
We should investigate!

PR #3451, opened 2023-08-05
fix: prevent double required config key in TCA

helsner:
Resolves: #3450

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -107,5 +107,8 @@
         "branch-alias": {
             "dev-main": "1.0-dev"
         }
+    },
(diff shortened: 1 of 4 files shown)


helsner:
hi @sabbelasichon - couldn't get behind this one.. but provided my broken tests
Detected while i did my presentation :D

mikestreety:
@sabbelasichon @helsner Hey both, can I ask what the reasoning was for adding `phpstan` as a conflict? I've just updated and have traced the addition back to here but couldn't find the reason for doing so

Issue #3179, opened 2022-10-02
Deprecation: #97126 - TCEforms removed in FlexForm


sabbelasichon:
Deprecation: #97126 - TCEforms removed in FlexForm

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-97126-TCEformsRemovedInFlexForm.html
.. include:: /Includes.rst.txt

==================================================
Deprecation: #97126 - TCEforms removed in FlexForm
==================================================

See :issue:`97126`

Description
===========

The `<TCEforms>` tag is no longer necessary (and allowed) in FlexForm
definitions. It was used to wrap TCA configuration and sheet
titles / descriptions. This wrapping must be omitted now.

Impact
======

Not omitting `TCEforms` will trigger a deprecation warning and log a message in
the deprecation log. An automatic migration is in place, which will eventually
be removed in upcoming TYPO3 versions.

Affected Installations
======================

*  All installations making use of FlexForm in their TCA. FlexForm definitions can
   be defined directly in TCA or can be external XML files.

*  Extensions, which extend the YAML configuration for forms with new fields.

Migration
=========

Omit the `<TCEforms>` tag in your FlexForm definition. The underlying
configuration moves one level up.

Before:

.. code-block:: xml

    <T3DataStructure>
        <ROOT>
            <TCEforms>
                <sheetTitle>sheet description 1</sheetTitle>
                <sheetDescription>
                    sheetDescription: Lorem ipsum dolor sit amet, consectetur adipiscing elit.
                </sheetDescription>
                <sheetShortDescr>
                    sheetShortDescr: Lorem ipsum dolor sit amet, consectetur adipiscing elit.
                </sheetShortDescr>
            </TCEforms>
            <type>array</type>
            <el>
                <input_1>
                    <TCEforms>
                        <label>input_1</label>
                        <config>
                            <type>input</type>
                        </config>
                    </TCEforms>
                </input_1>
            </el>
        </ROOT>
    </T3DataStructure>

After:

.. code-block:: xml

    <T3DataStructure>
        <ROOT>
            <sheetTitle>sheet description 1</sheetTitle>
            <sheetDescription>
                sheetDescription: Lorem ipsum dolor sit amet, consectetur adipiscing elit.
            </sheetDescription>
            <sheetShortDescr>
                sheetShortDescr: Lorem ipsum dolor sit amet, consectetur adipiscing elit.
            </sheetShortDescr>
[...]

PR #3460, opened 2023-10-07
[TASK] Add RemoveElementTceFormsRector

simonschaufi:
Resolves: #3179


simonschaufi:
The `yaml` for the Form framework should also be handled but with a different rector.

cweiske:
This has been removed with commit 7aac5b0234255a7bc2c03b1aa65375edd8021cd1.

PR #3462, opened 2023-10-10
[TASK] Simplify code generation command

simonschaufi:


simonschaufi:
@sabbelasichon I'm throwing a lot of code away and improve the whole code generation by using simple strings and making use of validators where they belong instead of within the normalizer where they were before. Are you ok with the change?

sabbelasichon:
@simonschaufi: I love Value Objects but in this case i don´t know why these value objects are not self validating. I.e. why is the Url Value Object not asserting that the parameter is a valid Url. Why does the description VO is not validating the min and max length? I think the value objects are not valuable in this case, but should be. Throw them away or refactor them to be self validating.

Issue #2861, opened 2022-05-06
Breaking: #96044 - Harden method signature of logicalAnd() and logicalOr()


sabbelasichon:
Breaking: #96044 - Harden method signature of logicalAnd() and logicalOr()

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Breaking-96044-HardenMethodSignatureOfLogicalAndAndLogicalOr.html
.. include:: /Includes.rst.txt

==========================================================================
Breaking: #96044 - Harden method signature of logicalAnd() and logicalOr()
==========================================================================

See :issue:`96044`

Description
===========

The method signature of :php:`\TYPO3\CMS\Extbase\Persistence\QueryInterface::logicalAnd()`
and :php:`\TYPO3\CMS\Extbase\Persistence\QueryInterface::logicalOr()` changed.
As a consequence the method signature of :php:`\TYPO3\CMS\Extbase\Persistence\Generic\Query::logicalAnd()`
and :php:`\TYPO3\CMS\Extbase\Persistence\Generic\Query::logicalOr()` changed as well.

Both methods do no longer accept an array as first parameter. Furthermore both
methods have two mandatory :php:`\TYPO3\CMS\Extbase\Persistence\Generic\Qom\ConstraintInterface`
parameters to form a logical conjunction on.

Both methods do indeed accept an infinite number of further constraints.

The :php:`logicalAnd()` method does now reliably return an instance of
:php:`\TYPO3\CMS\Extbase\Persistence\Generic\Qom\AndInterface` instance
while the :php:`logicalOr()` method returns a :php:`\TYPO3\CMS\Extbase\Persistence\Generic\Qom\OrInterface`
instance.

Impact
======

This change impacts all usages of said methods with

- either just one array parameter containing all constraints
- or passing just a single constraint

An array is no longer accepted as first parameter because it does not
guarantee the minimum number (2) of constraints is given.

Just one constraint is no longer accepted because in that case, the
incoming constraint would have simply been returned which is not compatible
with returning either a :php:`AndInterface` or a :php:`OrInterface` but just
a :php:`ConstraintInterface`.

Affected Installations
======================

All installations that passed all constraints as array or that passed just
one constraint.

Migration
=========

The migration is the same for :php:`logicalAnd()` and :php:`logicalOr()`
since their param signature is the same. The upcoming example will show a
migration for a :php:`logicalAnd()` call.

**Example**:

```php
[...]

PR #3640, opened 2023-10-19
[TASK] Harden method signature of logicalAnd() and logicalOr()

simonschaufi:
Resolves: #2861

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -79,12 +79,13 @@
         "contribute": [
             "@fix-rector",
             "@fix-style",
-            "@analyze"
+            "@analyze",
+            "@phpstan-generate-baseline"
         ],
         "check-style": "@php vendor/bin/ecs check --ansi",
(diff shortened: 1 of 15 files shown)


tests/Rector/v9/v0/QueryLogicalOrAndLogicalAndToArrayParameterRector/Fixture/query_logical_and_with_array_but_additional_parameter.php.inc
```diff
@@ -33,7 +33,7 @@ class ProductRepositoryLogicalAndButArrayAndAdditionalParameter extends Reposito
         $query->matching($query->logicalAnd([
                 $query->lessThan('foo', 1),
                 $query->lessThan('bar', 1),
-            ]
+            ], 'foo'
```
simonschaufi, line 36: This test changes its output because of the new Query stub. What shall I do here?

Issue #3351, opened 2023-03-06
Feature: #96874 - CKEditor 5


sabbelasichon:
Feature: #96874 - CKEditor 5

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Feature-96874-CKEditor5.html
.. include:: /Includes.rst.txt

.. _feature-96874-1664488673:

============================
Feature: #96874 - CKEditor 5
============================

See :issue:`96874`

Description
===========

TYPO3 v12 ships with CKEditor 5, a Rich-Text Editor to edit fields where
custom formatting for text with styling or links, or table formatting can be
achieved.

CKEditor 5 is a completely rewritten and new editor compared to CKEditor 4,
which was shipped since TYPO3 v8.

In general, most of the feature-set can be used in TYPO3 as before, with some
details kept in mind when upgrading.

Please read the documentation on the conceptual changes between CKEditor 4 and
CKEditor 5:

https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/migration-from-ckeditor-4.html

Impact
======

Next to plugins, which are not compatible anymore due to a completely different
model architecture, some configuration options have been modified or do not
apply anymore.

Most of the RTE configuration, which is done in TYPO3 in YAML preset files,
is migrated, however it is recommended to rewrite any custom configuration files
to become familiar with the CKEditor 5 API.

CSS Styling
-----------

CKEditor 5 does not load its editor in a specific iframe anymore. Especially
for adding custom styling and fonts, all CSS declarations now need to be prefixed
with ".ck-content". This can be achieved via SCSS, which TYPO3 natively
handles for custom CSS styles.
However, <body> tag CSS declarations won't work, as the `<body>` tag does not
apply to the editor HTML rendering anymore.

Configuration Options
---------------------

Some options have been adapted, which are rarely used, but now documented here:

*    editor.config.defaultContentLanguage is migrated to editor.config.language.content
*    editor.config.defaultLanguage is migrated to editor.config.language.ui

The following options are not needed anymore in CKEditor 5:

*    editor.config.uiColor
*    editor.config.removeDialogTabs
*    editor.config.entities_latin
*    editor.config.entities
*    editor.config.extraAllowedContent (covered via GeneralHTMLSupport plugin)

More migration options can be found here:
https://ckeditor.com/docs/ckeditor5/latest/installation/getting-started/migration-from-ckeditor-4.html

[...]

PR #3920, opened 2023-11-18
[FEATURE] Provide CKEditor5 config migration

simonschaufi:
Relates: #3351


simonschaufi:
@helsner you started this a long time ago. Do you want to finish it?

helsner:
@simonschaufi no, guess not
I talked to Benji Kott at the DevDays - the core wants to support v4 and v5 of the CKEditor.
So basically they're provide a proxy function thus making our migration not necessary IMHO

Issue #3940, opened 2023-12-22
Bug: SubstituteMethodRmFromListOfGeneralUtilityRector


franzholz:
Rector 0.17.0
ssch/typo3-rector 1.5.2
Typo3 11LTS

The replacement for rmFromList is broken and leads to a wrong code with an undefined variable $element. It loses the real variable  of the code.

example:

2) Classes/View/PhpWord.php:230

```
    ---------- begin diff ----------
@@ @@
             }
             $document->setValue('billing_' . $field, $value);
             if (!empty($value)) {
-                $missingFields = GeneralUtility::rmFromList($field, $missingFields);
+                $missingFields = implode(',', array_filter(explode(',', $missingFields), function ($item) use ($element) {
+                    return $element == $item;
+                }));
             }
             //         debug ($missingFields, '$missingFields abgezogen');
         }
    ----------- end diff -----------

```
Applied rules:
 * ReplaceInjectAnnotationWithMethodRector (https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.0/Breaking-90799-DependencyInjectionWithNonPublicPropertiesHasBeenRemoved.html)
 * SubstituteMethodRmFromListOfGeneralUtilityRector (https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/11.3/Deprecation-94311-DeprecatedGeneralUtilityrmFromList.html)
 * InjectAnnotationRector (https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/9.0/Feature-82869-ReplaceInjectWithTYPO3CMSExtbaseAnnotationInject.html)

It should have been like this instead:

```
+         $missingFields = implode(',', array_filter(explode(',', $missingFields), function ($item) use ($field) {
+             return $field == $item;
+         }));

```

PR #3941, opened 2023-12-22
[BUGFIX] Use arg as var name instead of fixed one

helsner:
Resolves: #3940


simonschaufi:
make sure the tests are greeen

simonschaufi (approved):

Issue #3600, opened 2023-10-13
Deprecation: #100071 - Magic repository findBy() methods


sabbelasichon:
Deprecation: #100071 - Magic repository findBy() methods

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.3/Deprecation-100071-MagicRepositoryFindByMethods.html
.. include:: /Includes.rst.txt

.. _deprecation-100071-1677853787:

========================================================
Deprecation: #100071 - Magic repository findBy() methods
========================================================

See :issue:`100071`

Description
===========

Extbase repositories come with a magic :php:`__call()` method to allow calling
the following methods without implementing them:

- :php:`findBy[PropertyName]($propertyValue)`
- :php:`findOneBy[PropertyName]($propertyValue)`
- :php:`countBy[PropertyName]($propertyValue)`

These have now been marked as deprecated, as they are "magic", meaning
that proper IDE support is not possible, and other PHP-related tool
functionality such as PhpStorm.

In addition, it is not possible for Extbase repositories
to build their own magic method functionality as the logic is already
in use.

Impact
======

As these methods are widely used in almost all Extbase-based extensions,
they are marked as deprecated in TYPO3 v12, but will only trigger a deprecation
notice in TYPO3 v13, as they will be removed in TYPO3 v14.

This way, migration towards the new API methods can be made without
pressure.

Affected installations
======================

All installations with third-party extensions that use those magic methods.

Migration
=========

A new set of methods without all the downsides have been added:

- :php:`findBy(array $criteria, ...): QueryResultInterface`
- :php:`findOneBy(array $criteria, ...):object|null`
- :php:`count(array $criteria, ...): int`

The naming of the methods follows those of `doctrine/orm` and only
:php:`count()` differs from the formerly :php:`countBy()`. While all magic
methods only allow for a single comparison (`propertyName` = `propertyValue`),
those methods allow for multiple comparisons, called constraints.

`findBy[PropertyName]($propertyValue)` can be replaced with a call to `findBy`:

..  code-block:: php

    $this->blogRepository->findBy(['propertyName' => $propertyValue]);

`findOneBy[PropertyName]($propertyValue)` can be replaced with a call to `findOneBy`:

..  code-block:: php

    $this->blogRepository->findOneBy(['propertyName' => $propertyValue]);

`countBy[PropertyName]($propertyValue)` can be replaced with a call to `count`:

..  code-block:: php

[...]

PR #3943, opened 2023-12-22
[FEATURE] Provide migration for extbase magic methods

helsner:
Resolves: #3600

### DIFF
--- a/src/Rector/v12/v3/typo3/MigrateMagicRepositoryMethodsRector.php
+++ b/src/Rector/v12/v3/typo3/MigrateMagicRepositoryMethodsRector.php
@@ -64,7 +64,7 @@ final class MigrateMagicRepositoryMethodsRector extends AbstractRector
 
         $methodReflection = $this->reflectionResolver->resolveMethodReflectionFromMethodCall($node);
         # Class reflection only works when the method exists - thus we don't migrate if it succeeds/is not null
-        if ($methodReflection !== null) {
+        if ($methodReflection instanceof MethodReflection) {
             return null;
         }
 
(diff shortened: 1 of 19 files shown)


helsner:
i could need some help her @simonschaufi 
I marked TODOs to check

src/Rector/v12/v3/typo3/MigrateMagicRepositoryMethodsRector.php
```diff
@@ -0,0 +1,108 @@
...
+    {
+        return [MethodCall::class];
+    }
+
+    /**
+     * @param Node\Expr\MethodCall $node
+     */
+    public function refactor(Node $node): ?Node
+    {
+        // TODO: this doesn't work with my ExampleRepo Stub
```
simonschaufi, line 36: no problem for me. I never return null here.

src/Rector/v12/v3/typo3/MigrateMagicRepositoryMethodsRector.php
```diff
@@ -0,0 +1,126 @@
...
+        if (! $this->nodeTypeResolver->isMethodStaticCallOrClassMethodObjectType(
+            $node,
+            new ObjectType('TYPO3\CMS\Extbase\Persistence\Repository')
+        )) {
+            return null;
+        }
+
+        // TODO: check if method already exists in class
+
+        $methodReflection = $this->reflectionResolver->resolveMethodReflectionFromMethodCall($node);
```
simonschaufi, line 56: I hope this helps you further. I found some nice code here: `\Rector\Core\NodeAnalyzer\DoctrineEntityAnalyzer::hasClassReflectionAttribute`

src/Rector/v12/v3/typo3/MigrateMagicRepositoryMethodsRector.php
helsner: I created a test stub class that extends the Repository class
With the stub class the check here does not work which is a problem i guess

Issue #3982, opened 2024-01-03
Move away from NodesToAddCollector


sabbelasichon:
The class NodesToAddCollector is deprecated. Return new nodes in refactor method instead.

sabbelasichon:
This is fixed as we have removed HardenMethodSignatureOfLogicalAndAndLogicalOrRector with https://github.com/sabbelasichon/typo3-rector/pull/4040

PR #3983, opened 2024-01-03
TASK: Return multiple nodes instead of using NodesToAddCollector

sabbelasichon:
Relates: #3982


sabbelasichon:
@simonschaufi We should either remove some complex rarely used rule which uses the NodesToAddCollector or refactor them. This can be quite cumbersome but we should move forward.

simonschaufi:
@sabbelasichon The title of the issue doesn't make it clear that it also removes some rules. Why did you actually remove them?

sabbelasichon:
Too old. Never used them. Never seen them in the wild tbh. Drop complexity. Focus on important thingd

simonschaufi:
ok but then we should also mark the commits somehow because we also don't really follow semver at the moment.

sabbelasichon:
I don't really care. Being pragmatic in that regard

PR #4015, opened 2024-01-09
TASK: Explicitly define arguments with tagged_iterator

sabbelasichon:
Relates: #4014


sabbelasichon:
Yes. It does not work and won't work

PR #4022, opened 2024-01-22
[TASK] Bump the PHPStan dependency

oliverklee:
Now the required PHPStan version matches the transitive dependency version from Rector. This should help reduce confusion about the PHPStan version that actually gets installed.


composer.json
```diff
@@ -28,7 +28,7 @@
         "helmich/typo3-typoscript-parser": "^2.4.1",
         "nette/utils": "^3.0 || ^4.0",
         "nikic/php-parser": "^4.17",
-        "phpstan/phpstan": "^1.9.7",
+        "phpstan/phpstan": "^1.10.15",
```
simonschaufi, line 31: I even have 1.10.27 installed. Can you confirm?

oliverklee: I have too. Then I'm going to bump to that version. :-) Hold on …

simonschaufi:
Thank you!

PR #4033, opened 2024-01-24
[TASK] Bump the dependencies

oliverklee:


sabbelasichon:
@oliverklee Thanks and fine for me for now. But i would refrain to bump the dependencies every week. This is not necessary IMHO.

PR #4034, opened 2024-01-24
[TASK] Narrow down the Symfony development dependencies

oliverklee:
The Symfony development dependencies do not need to be in sync with the TYPO3 Core Symfony dependencies as they are not installed in projects that have this package as a dependency. Instead, it suffices that the set of development dependencies is installable (and works) with the PHP versions we support.


sabbelasichon:
Same applies here. Thanks. But we should avoid to do it every week.

Issue #4013, opened 2024-01-09
Refactor AttributeDecoratorInterface


sabbelasichon:
AttributeDecoratorInterface is gone in rector 0.17.1

sabbelasichon:
Close as completed https://github.com/sabbelasichon/typo3-rector/pull/4041

Issue #3345, opened 2023-03-06
Feature: #96688 - Attributes for Extbase Annotations


sabbelasichon:
Feature: #96688 - Attributes for Extbase Annotations

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Feature-96688-AttributesForExtbaseAnnotations.html
.. include:: /Includes.rst.txt

.. _feature-96688:

====================================================
Feature: #96688 - Attributes for Extbase Annotations
====================================================

See :issue:`96688`

Description
===========

Since PHP 8, native attributes are supported. In comparison to doc comments, attributes have auto-completion,
are better readable and were "invented" for storing meta-information about properties.
For more info on attributes see https://stitcher.io/blog/attributes-in-php-8 and https://www.php.net/manual/en/language.attributes.overview.php

Extbase annotations are already nearly 1:1 translatable to attributes.

Impact
======

In addition to their usage as annotations, the following Extbase annotations have been enriched for usage as attributes:

..  code-block:: php

    @Extbase\ORM\Transient
    @Extbase\ORM\Cascade
    @Extbase\ORM\Lazy
    @Extbase\IgnoreValidation
    @Extbase\Validate

Examples
--------

Transient & Lazy
++++++++++++++++

Annotations::

    use TYPO3\CMS\Extbase\Annotation as Extbase;

    /**
     * @Extbase\ORM\Lazy()
     * @Extbase\ORM\Transient()
     */

Attributes::

    use TYPO3\CMS\Extbase\Annotation as Extbase;

    #[Extbase\ORM\Lazy()]
    #[Extbase\ORM\Transient()]

Cascade
+++++++

Annotation::

    /**
     * @Extbase\ORM\Cascade("remove")
     */

Attribute::

    #[Extbase\ORM\Cascade(['value' => 'remove'])]

Validate
++++++++

Annotations::

    /**
     * @Extbase\Validate("StringLength", options={"minimum": 1, "maximum": 10})
     * @Extbase\Validate("NotEmpty")
     * @Extbase\Validate("TYPO3.CMS.Extbase:NotEmpty")
     * @Extbase\Validate("TYPO3.CMS.Extbase.Tests.Unit.Reflection.Fixture:DummyValidator")
     * @Extbase\Validate("\TYPO3\CMS\Extbase\Validation\Validator\NotEmptyValidator")
[...]

PR #4041, opened 2024-01-24
[!!!] Removed annotations to attributes rules

sabbelasichon:
Relates: #3345, #4013


sabbelasichon:
The AttributeDecoratorInterface was removed in rector 0.17.1 We have to reimplement the Feature. It is a feature and not a hard removal. So i am fine with reimplementing when possible.

PR #4064, opened 2024-02-05
Release Version 2.x

sabbelasichon:


.github/workflows/code_analysis.yaml
```diff
@@ -4,7 +4,7 @@ on:
   pull_request: null
   push:
     branches:
-      - "main"
+      - "2.x"
```
simonschaufi, line 7: needs to be changed back to main

.github/workflows/docs.yaml
```diff
@@ -3,7 +3,7 @@ name: Docs
 on:
   push:
     branches:
-      - "main"
+      - "2.x"
```
simonschaufi, line 6: needs to be changed back to main

.github/workflows/docs.yaml
```diff
@@ -42,7 +42,7 @@ jobs:
         with:
           commit_message: '[DOCS] Generate documentation'
           commit_options: '--no-verify --signoff'
-          branch: "main"
+          branch: "2.x"
```
simonschaufi, line 45: needs to be changed back to main

simonschaufi (changes requested):

PR #4066, opened 2024-02-05
[BUGFIX] Make hidden dependency symfony/string explicit

sabbelasichon:


composer.json
```diff
@@ -35,6 +35,7 @@
         "symfony/finder": "^5.4 || ^6.4 || ^7.0",
         "symfony/polyfill-php80": "^1.28.0",
         "symfony/polyfill-php81": "^1.28.0",
+        "symfony/string": "^5.0 || ^6.0 || ^7.0",
```
oliverklee, line 38: This probably should be `^5.4 || ^6.4 || 7.0` to be in line with the other Symfony dependencies.

PR #4118, opened 2024-02-13
[BUGFIX] Fix default rector.php.dist

simonschaufi:


templates/rector.php.dist
```diff
@@ -16,12 +15,14 @@ return RectorConfig::configure()
         ExtEmConfRector::ADDITIONAL_VALUES_TO_BE_REMOVED => []
     ])
     ->withPaths([
-        __PATHS__
+        __DIR__
```
sabbelasichon, line 18: This has to be __PATHS__ it will be substituted by a str_replace with some autodetection of directories of PHP files. Have a look at: https://github.com/sabbelasichon/typo3-rector/blob/main/bin/typo3-init.php#L44

sabbelasichon:
@simonschaufi Thanks, everything else makes totally sense. Thanks for fixing it.

templates/rector.php.dist
simonschaufi: ah ok. thanx

Issue #4137, opened 2024-03-05
Falsely removal of eval on datetime migration


helsner:
The eval => required is thrown away instead of migrated or at least left within the config.
Guess this happens with the MigrateInputDateTimeRector
TCA:
```
'config' => [
-                'type' => 'input',
-                'renderType' => 'inputDateTime',
+                'type' => 'datetime',
                 'dbType' => 'date',
-                'eval' => 'date,required',
+                'format' => 'date',
             ],
```

PR #4138, opened 2024-03-05
[BUGFIX] Fix wrong eval removal on TCA datetime migration

helsner:
Resolves: #4137


helsner:
have to fix this, right now its just the broken test

helsner:
The issue is that we cannot remove the required eval within this rector.
We have one for the Datetime Migration and one for the required => true migration
But if the Datetime Migration runs before the eval => required to required => true one, then the value is getting kicked.
eval => null has the same behavior, but we handle it properly and the Nullable TCA Rector does its job afterwards

simonschaufi (approved):

PR #4140, opened 2024-03-06
Leave FQCN unchanged

mbrodala:
Existing FQCN should not be changed as drive-by-change but willingly and explicitly in each project instead. This also reduces the diff in many cases and thus simplifies selectively committing Rector changes.

See https://github.com/sabbelasichon/typo3-rector/discussions/2972


mbrodala:
Tests need adoption but first I'd like to know if this change would be accepted ...

sabbelasichon:
@mbrodala I would love to see this merged. Are you willing to move this PR forward?

mbrodala:
Willing definitely but right now my time is limited so I won't be able to continue here for the time being. If you want to clean up you can close this and I'll recreate it once I have more time.

sabbelasichon:
Have a look at #4230

Issue #3699, opened 2023-11-03
Deprecation: #101554 - Obsolete TCA MM_hasUidField


simonschaufi:
Deprecation: #101554 - Obsolete TCA MM_hasUidField

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/13.0/Deprecation-101554-ObsoleteTCAMM_hasUidField.html
.. include:: /Includes.rst.txt

.. _deprecation-101554-1691480627:

==================================================
Deprecation: #101554 - Obsolete TCA MM_hasUidField
==================================================

See :issue:`101554`

Description
===========

When configuring :php:`MM` relations in TCA, the field :php:`MM_hasUidField`
has been obsoleted: A :sql:`uid` column is only needed when :php:`multiple`
is set to true - when a record is allowed to be selected multiple times in
a relation. In this case, the :sql:`uid` field is added automatically by the
database analyzer.

Impact
======

The TCA configuration option :php:`MM_hasUidField` is obsolete and can be removed.

The :php:`TcaMigration` performed during TCA warmup will automatically remove the
option and creates according log entries if needed.

Affected installations
======================

Instances with extensions using MM relations may be affected.

Migration
=========

Remove all occurrences of php:`MM_hasUidField` from TCA. The :sql:`uid` column
is added as primary key automatically if :php:`multiple = true` is set, otherwise
a combined primary key of fields :sql:`uid_local`, :sql:`uid_foreign` plus
eventually :sql:`tablenames` and :sql:`fieldname` is used.

.. index:: TCA, NotScanned, ext:core

PR #4141, opened 2024-03-06
[FEATURE] Add TCA Rule: remove mm_hasUidField config

126016:
Resolves: #3699


tests/Rector/v13/v0/RemoveMmHasUidFieldRector/Fixture/fixture.php.inc
```diff
@@ -0,0 +1,32 @@
+<?php
+
+namespace Ssch\TYPO3Rector\Tests\Rector\v13\v0\RemoveMmHasUidFieldRector\Fixture;
+
+return [
+    'ctrl' => [],
```
sabbelasichon, line 6: Try it without ctrl for single TCA column definitions in Overrides

126016: We add a new fixture with the TCA globals example. It works, but the control is mandatory.

Issue #3613, opened 2023-10-13
Deprecation: #100405 - Property TypoScriptFrontendController->type


sabbelasichon:
Deprecation: #100405 - Property TypoScriptFrontendController->type

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.4/Deprecation-100405-PropertyTypoScriptFrontendController-type.html
.. include:: /Includes.rst.txt

.. _deprecation-100405-1680520177:

==================================================================
Deprecation: #100405 - Property TypoScriptFrontendController->type
==================================================================

See :issue:`100405`

Description
===========

The public property :php:`type` of the main class in TYPO3 frontend
:php:`\TYPO3\CMS\Frontend\Controller\TypoScriptFrontendController` has been
marked as internal, as it should not be used outside of this PHP class anymore
in the future.

This is part of the overall part to reduce dependencies on this PHP class, as
it is not always available in TYPO3 frontend.

Impact
======

Accessing this property will trigger a PHP deprecation notice. Accessing this
property might also happen via TypoScript and TypoScript conditions.

Affected installations
======================

TYPO3 installations using this property on checking various typeNum settings
from TypoScript.

Migration
=========

When using this property in PHP code via :php:`$GLOBALS['TSFE']->type`, it is
recommended to move to the PSR-7 request via
:php:`$request->getAttribute('routing')->getPageType()`, which is the property
of the :php:`PageArguments` object, as a result of the :php:`GET` parameter
:php:`type`, or `$GLOBALS['TSFE']->getPageArguments()->getPageType()` if
the request object is not available.

Within TypoScript, conditions and getData properties need to be adapted:

..  code-block:: typoscript

    # Before
    [getTSFE() && getTSFE().type == 13]

    # After
    [request.getPageArguments()?.getPageType() == 13]

In TypoScript getData attributes:

..  code-block:: typoscript

    # Before
    page.10.data = TSFE:type

    # After
    page.10.data = request:routing|pageType

.. index:: Frontend, TypoScript, FullyScanned, ext:frontend

PR #4164, opened 2024-03-08
[FEATURE] Add MigrateTypoScriptFrontendControllerTypeRector

simonschaufi:
Resolves: #3613

### DIFF
--- a/rules/TYPO312/v4/MigrateTypoScriptFrontendControllerTypeRector.php
+++ b/rules/TYPO312/v4/MigrateTypoScriptFrontendControllerTypeRector.php
@@ -1,0 +1,109 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\TYPO312\v4;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\MethodCall;
+use PhpParser\Node\Expr\PropertyFetch;
+use PHPStan\Type\ObjectType;
+use Rector\Rector\AbstractRector;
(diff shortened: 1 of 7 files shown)


rules/TYPO312/v4/MigrateTypoScriptFrontendControllerTypeRector.php
```diff
@@ -0,0 +1,105 @@
...
+    public function getNodeTypes(): array
+    {
+        return [PropertyFetch::class];
+    }
+
+    /**
+     * @param PropertyFetch $node
+     * @return Node|null
+     */
+    public function refactor(Node $node)
```
sabbelasichon, line 67:
```suggestion
    public function refactor(Node $node): ?Node
```

rules/TYPO312/v4/MigrateTypoScriptFrontendControllerTypeRector.php
```diff
@@ -0,0 +1,105 @@
...
+        if (! $this->isNames($node->name, [self::TYPE])) {
+            return null;
+        }
+
+        return $this->createGetPageTypeMethodCall(
+            $this->createTSFEGetPageArgumentsMethodCall()
+        );
+    }
+
+    private function shouldSkip(PropertyFetch $propertyFetch): bool
```
sabbelasichon, line 82: This is not the only case. You could also have cases where some is using a wrapper method. You should also check for the method call on TypoScriptFrontendController

```php
private function getTypoScriptFrontendController(): TypoScriptFrontendController
{
    return $GLOBALS['TSFE'];
}

Issue #3621, opened 2023-10-13
Deprecation: #100596 - GeneralUtility::_GET()


sabbelasichon:
Deprecation: #100596 - GeneralUtility::_GET()

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.4/Deprecation-100596-GeneralUtility_GET.html
.. include:: /Includes.rst.txt

.. _deprecation-100596-1681478199:

=============================================
Deprecation: #100596 - GeneralUtility::_GET()
=============================================

See :issue:`100596`

Description
===========

The method :php:`\TYPO3\CMS\Core\Utility\GeneralUtility::_GET()` has
been marked as deprecated and should not be used any longer.

Modern code should access GET and POST data from the PSR-7 :php:`ServerRequestInterface`,
and should avoid accessing superglobals :php:`$_GET` directly. This also avoids
future side-effects when using sub-requests. Some :php:`GeneralUtility` related
helper methods like :php:`_GET()` violate this, using them is considered a technical
debt. They are being phased out.

Impact
======

Calling the method from PHP code will log a PHP deprecation level entry,
the method will be removed with TYPO3 v13.

Affected installations
======================

TYPO3 installations with third-party extensions using :php:`GeneralUtility::_GET()`
are affected, typically in TYPO3 installations which
have been migrated to the latest TYPO3 Core versions and
haven't been adapted properly yet.

The extension scanner will find usages with a strong match.

Migration
=========

:php:`GeneralUtility::_GET()` is a helper method that retrieves
incoming HTTP `GET` query arguments and returns the value.

The same result can be achieved by retrieving arguments from the request object.
An instance of the PSR-7 :php:`ServerRequestInterface` is handed over to
controllers by TYPO3 Core's PSR-15 :php:`\TYPO3\CMS\Core\Http\RequestHandlerInterface`
and middleware implementations, and is available in various related scopes
like the frontend :php:`\TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer`.

Typical code:

..  code-block:: php

    use TYPO3\CMS\Core\Utility\GeneralUtility;

    // Before
    $value = GeneralUtility::_GET('tx_scheduler');

    // After
    $value = $request->getQueryParams()['tx_scheduler']) ?? null;

.. index:: PHP-API, FullyScanned, ext:core

Issue #3293, opened 2023-03-06
Deprecation: #99633 - GeneralUtility::_POST()


sabbelasichon:
Deprecation: #99633 - GeneralUtility::_POST()

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Deprecation-99633-GeneralUtilityPOST.html
.. include:: /Includes.rst.txt

.. _deprecation-99633-1674121794:

=============================================
Deprecation: #99633 - GeneralUtility::_POST()
=============================================

See :issue:`99633`

Description
===========

The method :php:`\TYPO3\CMS\Backend\Utility\GeneralUtility::_POST()` has
been marked deprecated and should not be used any longer.

Modern code should access GET and POST data from the PSR-7
:php:`\Psr\Http\Message\ServerRequestInterface`, and should avoid accessing
super-globals :php:`$_GET` and :php:`$_POST`
directly. This will avoid future side-effects when using sub-requests. Some
:php:`GeneralUtility` related helper methods like :php:`_POST()` violate this,
using them is considered a technical debt. They are being phased out.

Impact
======

Calling the method from PHP code will trigger a PHP deprecation notice.

Affected installations
======================

TYPO3 installations with third-party extensions using :php:`GeneralUtility::_POST()`
are affected. This typically occurs in TYPO3 installations which
have been migrated to latest TYPO3 Core versions and
haven't been adapted properly yet.

The extension scanner will find usages with a strong match.

Migration
=========

:php:`GeneralUtility::_POST()` is a helper method that retrieves
incoming HTTP body parameters / `POST` parameters and returns the value.

The same result can be achieved by retrieving arguments from the request object.
An instance of the PSR-7 :php:`ServerRequestInterface` is handed over to
controllers by TYPO3 Core's PSR-15 :php:`\TYPO3\CMS\Core\Http\RequestHandlerInterface`
and middleware implementations, and is available in various related scopes
like the frontend :php:`\TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer`.

Typical code:

..  code-block:: php

    use TYPO3\CMS\Backend\Utility\GeneralUtility;

    // Before
    $value = GeneralUtility::_POST('tx_scheduler');

    // After
    $value = $request->getParsedBody()['tx_scheduler']);

.. index:: Backend, PHP-API, FullyScanned, ext:backend

PR #4181, opened 2024-03-21
[FEATURE] Use PSR-7 instead of GeneralUtility::_GET  

sabbelasichon:
Resolves: #3621

### DIFF
--- a/rules/TYPO312/v4/UseServerRequestInsteadOfGeneralUtilityGetRector.php
+++ b/rules/TYPO312/v4/UseServerRequestInsteadOfGeneralUtilityGetRector.php
@@ -108,7 +108,7 @@ CODE_SAMPLE
 
         $classReflection = $scope->getClassReflection();
 
-        if ($classReflection !== null && $classReflection->isSubclassOf(
+        if ($classReflection instanceof \PHPStan\Reflection\ClassReflection && $classReflection->isSubclassOf(
             'TYPO3\CMS\Extbase\Mvc\Controller\ActionController'
         )) {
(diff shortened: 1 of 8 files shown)


rules/TYPO312/v4/UseServerRequestInsteadOfGeneralUtilityGetRector.php
```diff
@@ -0,0 +1,115 @@
...
+        return new RuleDefinition('Use PSR-7 ServerRequest instead of GeneralUtility::_GET()', [
+            new CodeSample(
+                <<<'CODE_SAMPLE'
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+$value = GeneralUtility::_GET('tx_scheduler');
+CODE_SAMPLE
+                ,
+                <<<'CODE_SAMPLE'
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+$value = $request->getQueryParams()['tx_scheduler'] ?? null;
```
simonschaufi, line 51:
```suggestion
$value = $GLOBALS['TYPO3_REQUEST']->getQueryParams()['tx_scheduler'] ?? null;
```

simonschaufi (changes requested):

simonschaufi:
> @simonschaufi What should we do if _GET parameter is null? Should we do nothing or retrieve getQueryParams as a whole?

I would leave it up to the developer to fix his code but that is not our responsibility.

simonschaufi (approved):

simonschaufi:
now you can basically copy the code and do the same thing for https://github.com/sabbelasichon/typo3-rector/issues/3293

PR #4199, opened 2024-03-26
[TASK] Include Extension root files for Rector to parse

simonschaufi:
Some rules like the ExtEmConfRector need the root files to work


simonschaufi:
@sabbelasichon some rules like the ExtEmConfRector need the root files to work

Issue #3292, opened 2023-03-06
Deprecation: #99615 - GeneralUtility::_GPmerged()


sabbelasichon:
Deprecation: #99615 - GeneralUtility::_GPmerged()

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/12.2/Deprecation-99615-GeneralUtilityGPMerged.html#deprecation-99615-generalutility-gpmerged

.. include:: /Includes.rst.txt

.. _deprecation-99615-1674056024:

=================================================
Deprecation: #99615 - GeneralUtility::_GPmerged()
=================================================

See :issue:`99615`

Description
===========

The method :php:`\TYPO3\CMS\Backend\Utility\GeneralUtility::_GPmerged()` has
been marked as deprecated and should not be used any longer.

Modern code should access `GET` and `POST` data from the PSR-7
:php:`\Psr\Http\Message\ServerRequestInterface`, and should avoid accessing
super-globals :php:`$_GET` and :php:`$_POST` directly. This helps creating
controller classes with a clean architecture. Some
:php:`\TYPO3\CMS\Core\Utility\GeneralUtility` related helper methods like
:php:`_GPmerged()` violate this, using them is considered a technical debt.
They are being phased out.

Impact
======

Calling the method will raise a deprecation level log error and will
stop working with TYPO3 v13.

Affected installations
======================

Instances with extensions using :php:`GeneralUtility::_GPmerged()` are affected.
The extension scanner will find usages with a strong match.

Migration
=========

:php:`GeneralUtility::_GPmerged()` is a helper method that retrieves
request parameters and returns the value, while `POST` parameters take
precedence over `GET` parameters, if both exist.

The same result can be achieved by retrieving arguments from the request object.
An instance of the PSR-7 :php:`ServerRequestInterface` is handed over to
controllers by TYPO3 Core's PSR-15 :php:`\TYPO3\CMS\Core\Http\RequestHandlerInterface`
and middleware implementations, and is available in various related scopes
like the frontend :php:`\TYPO3\CMS\Frontend\ContentObject\ContentObjectRenderer`.

Typical code:

..  code-block:: php

    use TYPO3\CMS\Backend\Utility\GeneralUtility;
    use TYPO3\CMS\Core\Utility\ArrayUtility;

    // Before
    $getMergedWithPost = GeneralUtility::_GPmerged('tx_scheduler');

    // After
[...]

PR #4213, opened 2024-04-04
[FEATURE] Add MigrateGeneralUtilityGPMergedRector

simonschaufi:
Resolves: #3292

### DIFF
--- a/rules/TYPO312/v2/MigrateGeneralUtilityGPMergedRector.php
+++ b/rules/TYPO312/v2/MigrateGeneralUtilityGPMergedRector.php
@@ -1,0 +1,175 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\TYPO312\v2;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\ArrayDimFetch;
+use PhpParser\Node\Expr\Assign;
+use PhpParser\Node\Expr\PropertyFetch;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Expr\StaticPropertyFetch;
+use PhpParser\Node\Expr\Variable;
+use PhpParser\Node\Stmt\Expression;
+use PHPStan\Analyser\Scope;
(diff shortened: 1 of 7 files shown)


simonschaufi:
My first rule with Expressions! :partying_face:

rules/TYPO312/v2/MigrateGeneralUtilityGPMergedRector.php
```diff
@@ -0,0 +1,171 @@
...
+
+        $staticCall = $assignNode->expr;
+
+        if (! $staticCall instanceof StaticCall) {
+            return null;
+        }
+
+        if (! $this->nodeTypeResolver->isMethodStaticCallOrClassMethodObjectType(
+            $staticCall,
+            new ObjectType(GeneralUtility::class)
```
sabbelasichon, line 115: for convention use string instead of ::class here

rules/TYPO312/v2/MigrateGeneralUtilityGPMergedRector.php
```diff
@@ -0,0 +1,171 @@
...
+     */
+    public function refactorWithScope(Node $node, Scope $scope): ?array
+    {
+        $assignNode = $node->expr;
+        if (! $assignNode instanceof Assign) {
+            return null;
+        }
+
+        /** @var Variable $variable */
+        $variable = $assignNode->var;
```
sabbelasichon, line 105: Do not expect var to be a variable. I would prefer to use a strict type check with an early return statement

sabbelasichon:
@simonschaufi I would also add a test fixture with a different static call method of GeneralUtility and expect nothing to be happen.

rules/TYPO312/v2/MigrateGeneralUtilityGPMergedRector.php
simonschaufi: we could write a custom rector rule for this to write class names as string ìf they start with `TYPO3` :smile:

Issue #4215, opened 2024-04-05
ExtbaseControllerActionsMustReturnResponseInterfaceRector doesn't consider existing RedirectResponse


simonschaufi:
# Bug Report

```diff
         return new RedirectResponse('...');
+        return $this->htmlResponse();
     }
```

## Expected Behaviour

do nothing

PR #4221, opened 2024-04-09
[BUGFIX] Do not add return statement if already of type ResponseInterface

sabbelasichon:
Resolves: #4215


simonschaufi:
thanks for the fix!

Issue #4223, opened 2024-04-10
Improve GeneralUtility::_POST migration


simonschaufi:
If the current code is already with `??`, it could be considered imho.

Current code:

```php
private function initArguments(): void
{
    $this->arguments = GeneralUtility::_POST('variable') ?? [];
}
```

current migration:

```php
private function initArguments(): void
{
    $this->arguments = ($GLOBALS['TYPO3_REQUEST']->getParsedBody()['variable'] ?? null) ?? [];
}
```

better would be:

```php
private function initArguments(): void
{
    $this->arguments = $GLOBALS['TYPO3_REQUEST']->getParsedBody()['variable'] ?? [];
}
```

See also #4224

PR #4227, opened 2024-04-11
[BUGFIX] Consider existing coalesce for _GET and _POST

simonschaufi:
Resolves: #4223


simonschaufi:
@sabbelasichon I'm not happy with the globalsToPsr7ServerRequestFactory at the moment because it does a lot more now than it should. How would you refactor this class to handle the global variables as well?

PR #4230, opened 2024-04-15
[!!!][TASK] Do not configure import names per default

sabbelasichon:


simonschaufi (approved):

mbrodala:
Nice one, way better than having to change all tests. Thanks a lot!

Issue #3347, opened 2023-03-06
Feature: #96733 - New backend module registration API


sabbelasichon:
Feature: #96733 - New backend module registration API

https://docs.typo3.org/c/typo3/cms-core/master/en-us/Changelog/12.0/Feature-96733-NewBackendModuleRegistrationAPI.html
.. include:: /Includes.rst.txt

.. _feature-96733:

=====================================================
Feature: #96733 - New backend module registration API
=====================================================

See :issue:`96733`

Description
===========

The registration and usage of backend modules was previously based
on the global array :php:`$TBE_MODULES`. This however had a couple
of drawbacks, e.g. module registration could be changed at runtime,
which had been resolved by introducing a new registration API.

Therefore, instead of using the :php:`ExtensionManagementUtility::addModule()`
and :php:`ExtensionUtility::registerModule()` (Extbase) API methods in
:file:`ext_tables.php` files, the configuration is now placed in the
dedicated :file:`Configuration/Backend/Modules.php` configuration file.

Those files are then read and processed when building the container. This
means the state is fixed and can't be changed at runtime. This approach
follows the general Core strategy (see e.g. :doc:`Icons.php <../11.4/Feature-94692-RegisteringIconsViaServiceContainer>`),
since it highly improves the loading speed of every request as the
registration can be handled at once and cached during warmup of the
Core caches. Besides caching, this will also allow additional features
in the future, which were blocked due to the loose state.

Previous configuration in :file:`ext_tables.php`:

```php
    ExtensionManagementUtility::addModule(
        'web',
        'example',
        'top',
        '',
        [
            'routeTarget' => MyExampleModuleController::class . '::handleRequest',
            'name' => 'web_example',
            'access' => 'admin',
            'workspaces' => 'online',
            'iconIdentifier' => 'module-example',
            'labels' => 'LLL:EXT:example/Resources/Private/Language/locallang_mod.xlf',
            'navigationComponentId' => 'TYPO3/CMS/Backend/PageTree/PageTreeElement',
        ]
    );

    ExtensionUtility::registerModule(
        'Extkey',
        'web',
        'example',
        'after:info',
        [
            MyExtbaseExampleModuleController::class => 'list, detail',
        ],
        [
            'access' => 'admin',
            'workspaces' => 'online',
            'iconIdentifier' => 'module-example',
            'labels' => 'LLL:EXT:extkey/Resources/Private/Language/locallang_mod.xlf',
[...]

PR #4243, opened 2024-04-22
[FEATURE] Add MigrateBackendModuleRegistrationRector

simonschaufi:
Resolves: #3347

### DIFF
--- a/rules/TYPO312/v0/MigrateBackendModuleRegistrationRector.php
+++ b/rules/TYPO312/v0/MigrateBackendModuleRegistrationRector.php
@@ -1,0 +1,581 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\TYPO312\v0;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\StaticCall;
+use PhpParser\Node\Stmt\Expression;
+use PhpParser\Node\Stmt\Return_;
+use PhpParser\NodeTraverser;
+use PHPStan\Type\ObjectType;
+use Rector\Exception\ShouldNotHappenException;
+use Rector\NodeTypeResolver\NodeScopeAndMetadataDecorator;
+use Rector\PhpParser\Node\CustomNode\FileWithoutNamespace;
+use Rector\PhpParser\Node\Value\ValueResolver;
+use Rector\PhpParser\Parser\RectorParser;
+use Rector\PhpParser\Printer\BetterStandardPrinter;
+use Rector\Rector\AbstractRector;
+use Rector\ValueObject\Application\File;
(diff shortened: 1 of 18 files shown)


simonschaufi:
TODO:
- Read existing array
- Maybe parse module call into array variables?
- Render array and write it into the Modules.php file

simonschaufi:
@samsonasik Dear Abdul, this is probably the most difficult Rector rule ever you have seen so far because I not only need to change code within a file but across different files. Creating files and writing to it is no problem for me as we have several other rules where we do this already. The big challenge here is to parse an existing php array and add a new item into it.

My initial thought was that i just create the new php array in pure php and then let the php parser do its thing and then I just merge the 2 nodes together but this isn't working because of some internal things I don't understand. For some reason the old nodes get duplicated even though the nodes are properly merged when you debug the node tree. My guess is that is has to do with the attributes that get attached to each node but I'm not sure.

My second try was to create the ArrayItem by hand but then I have the issue that the written output is just a one liner (see the 2 versions below the comment `Merge the arrays`).

Maybe you have some deeper understanding on what is going on internally in the php parser and have an idea that could solve my issue here?

rules/TYPO312/v0/MigrateBackendModuleRegistrationRector.php
```diff
@@ -0,0 +1,499 @@
...
+        /** @var Array_ $array */
+        $array = $return->expr;
+
+        return $array;
+    }
+
+    private function printFile(File $file, string $filePath) : void
+    {
+        // only save to string first, no need to print to file when not needed
+        $newContent = $this->formatPreservingPrinter->printParsedStmstAndTokensToString($file);
```
samsonasik, line 493: this removed at 

- https://github.com/rectorphp/rector-src/pull/6036

use directly from `betterStandardPrinter` instead:

```php
        $newContent = $this->betterStandardPrinter->printFormatPreserving(
             $file->getNewStmts(),
            $file->getOldStmts(),
            $file->getOldTokens()
       );
```

simonschaufi:
@samsonasik thank you very much for looking into it bit even with this change the problem still exists that the output is currently still the duplicate of `tools_toolsmaintenance` instead of the added `tools_toolssettings`

samsonasik:
the array structure may be appended, the structure may be need to be unset and append by index, like using `array_splice()`...

simonschaufi:
I'm merging the two arrays here: https://github.com/sabbelasichon/typo3-rector/pull/4243/files#diff-6659bbad2dc4a8d574769de61529b6e6d9387889d39813bbe2f246c873ba3bfeR461

When you inspect the code afterwards, it looks correct but the generated code is not correct and that is strange. Is there somebody in the Rector team who has more understanding on what is going on under the hood?

simonschaufi:
ping @samsonasik @TomasVotruba

Issue #4256, opened 2024-04-29
[BUG] Missing negation on SubstituteConstantsModeAndRequestTypeRector


helsner:
### Minimal PHP Code Causing Issue

```php
if (TYPO3_MODE !== 'FE') {
```

### Applied rules

SubstituteConstantsModeAndRequestTypeRector

### Expected Behaviour

```php
if (!ApplicationType::fromRequest($GLOBALS['TYPO3_REQUEST'])->isFrontend()) {
  // Do something
}
```

### Package Version

main

### PHP Version

8.1

### TYPO3 Version

v11

PR #4278, opened 2024-05-27
[BUGFIX] Add missing negate operator

helsner:
Resolves: #4256


rules/TYPO312/v0/MoveAllowTableOnStandardPagesToTCAConfigurationRector.php
```diff
@@ -90,13 +90,15 @@ public function refactor(Node $node)
         $tableNames = $this->valueResolver->getValue($tableArgument);
 
         foreach (explode(',', $tableNames) as $tableName) {
-            if (!$tableName) {
+            if ($tableName === '' || $tableName === '0') {
```
simonschaufi, line 93: these changes should be reverted and a table name should never be 0

Issue #4285, opened 2024-07-02
[BUG]: Class "Rector\Symfony\NodeAnalyzer\Command\SetAliasesMethodCallExtractor" does not exist


mbrodala:
### Minimal PHP Code Causing Issue

```php
<?php

declare(strict_types=1);

/*
 * This file is part of the package t3g/file_variants.
 *
 * For the full copyright and license information, please read the
 * LICENSE file that was distributed with this source code.
 */

use Rector\Config\RectorConfig;
use Ssch\TYPO3Rector\Set\Typo3SetList;
use Ssch\TYPO3Rector\TYPO312\v3\MigrateItemsIndexedKeysToAssociativeRector;

return RectorConfig::configure()
    ->withPaths([
        // ...
    ])
    ->withSets([
        Typo3SetList::TYPO3_12,
    ])
;
```

### Applied rules

None, Rector fails to run:

```
$ rector --config=Build/rector.php --dry-run --no-progress-bar
PHP Fatal error:  Uncaught ReflectionException: Class "Rector\Symfony\NodeAnalyzer\Command\SetAliasesMethodCallExtractor" does not exist in /.../vendor/rector/rector/vendor/illuminate/container/Container.php:782
Stack trace:
#0 /.../vendor/rector/rector/vendor/illuminate/container/Container.php(782): ReflectionClass->__construct('Rector\\Symfony\\...')
#1 /.../vendor/rector/rector/vendor/illuminate/container/Container.php(681): RectorPrefix202407\Illuminate\Container\Container->build('Rector\\Symfony\\...')
#2 /.../vendor/rector/rector/vendor/illuminate/container/Container.php(628): RectorPrefix202407\Illuminate\Container\Container->resolve('Rector\\Symfony\\...', Array)
#3 /.../vendor/rector/rector/vendor/illuminate/container/Container.php(906): RectorPrefix202407\Illuminate\Container\Container->make('Rector\\Symfony\\...')
#4 /.../vendor/rector/rector/vendor/illuminate/container/Container.php(836): RectorPrefix202407\Illuminate\Container\Container->resolveClass(Object(ReflectionParameter))
#5 /.../vendor/rector/rector/vendor/illuminate/container/Container.php(806): RectorPrefix202407\Illuminate\Container\Container->resolveDependencies(Array)
[...]

PR #4287, opened 2024-07-02
[BUGFIX] Fix missing class

sabbelasichon:
Resolves: #4285


mbrodala:
Thanks for the [quick fix](https://github.com/TYPO3GmbH/ext-file-variants/actions/runs/9753757743/job/26941232034)!

Issue #4291, opened 2024-07-18
[BUG]: MoveExtensionManagementUtilityAddStaticFileIntoTCAOverridesRector puts generated file in wrong folder


simonschaufi:
### Minimal PHP Code Causing Issue

```php
\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::addStaticFile('ext_name', 'Configuration/TypoScript', 'Title');
```

### Applied rules

MoveExtensionManagementUtilityAddStaticFileIntoTCAOverridesRector

### Expected Behaviour

New file under packages/ext_name/...

Instead new file under home/user/project/packages/ext_name/...

### Package Version

latest

### PHP Version

8.1

### TYPO3 Version

12.4

PR #4277, opened 2024-05-24
[BUGFIX] fix FilesystemInterface prefix

Kanti:
given:
file in `/app/src/extensions/aus_project/ext_tables.php` with content:
````php
<?php
\TYPO3\CMS\Core\Utility\ExtensionManagementUtility::allowTableOnStandardPages('tx_ausproject_domain_model_slide,tx_ausproject_domain_model_teaser');
````

running `rector process --debug src/extensions/aus_project/ext_tables.php` in /app directory.

Producess:
````
[ERROR] Could not process "src/extensions/aus_project/ext_tables.php" file, due to:
         "System error: "Unable to create a directory at /src/extensions/aus_project/Configuration/TCA/Overrides.
         mkdir(): Permission denied"

         Stack trace:
         #0 vendor/league/flysystem/src/Local/LocalFilesystemAdapter.php(297):
         League\Flysystem\UnableToCreateDirectory::atLocation('/src/extensions...', 'mkdir(): Permis...')
         #1 vendor/league/flysystem/src/Local/LocalFilesystemAdapter.php(116):
         League\Flysystem\Local\LocalFilesystemAdapter->ensureDirectoryExists('/src/extensions...', 448)
         #2 vendor/league/flysystem/src/Local/LocalFilesystemAdapter.php(102):
         League\Flysystem\Local\LocalFilesystemAdapter->writeToFile('src/extensions/...', '<?php\n\n// Inclu...',
         Object(League\Flysystem\Config))
         #3 vendor/league/flysystem/src/Filesystem.php(41):
         League\Flysystem\Local\LocalFilesystemAdapter->write('src/extensions/...', '<?php\n\n// Inclu...',
         Object(League\Flysystem\Config))
         #4 vendor/ssch/typo3-rector/src/Filesystem/FlysystemFilesystem.php(24):
         League\Flysystem\Filesystem->write('src/extensions/...', '<?php\n\n// Inclu...')
         #5
         vendor/ssch/typo3-rector/rules/CodeQuality/General/MoveExtensionManagementUtilityAddStaticFileIntoTCAOverridesR
         ector.php(121): Ssch\TYPO3Rector\Filesystem\FlysystemFilesystem->write('src/extensions/...', '<?php\n\n//
         Inclu...')
         #6 vendor/rector/rector/src/Rector/AbstractRector.php(136):
         Ssch\TYPO3Rector\CodeQuality\General\MoveExtensionManagementUtilityAddStaticFileIntoTCAOverridesRector->refact
[...]

PR #4292, opened 2024-07-18
[BUGFIX] Use absolute file path without project prefix

simonschaufi:
Resolves: #4291

### DIFF
--- a/config/config.php
+++ b/config/config.php
@@ -31,12 +31,7 @@ return static function (RectorConfig $rectorConfig): void {
         if (StaticPHPUnitEnvironment::isPHPUnitRun() || $isDryRun) {
             $adapter = new InMemoryFilesystemAdapter();
         } else {
-            $cwd = getcwd();
-            if (! $cwd) {
-                throw new \Exception('could not find current working directory');
-            }
-
-            $adapter = new LocalFilesystemAdapter($cwd);
+            $adapter = new LocalFilesystemAdapter('/');
         }
 
         return new FlysystemFilesystem(new Filesystem($adapter));


Kanti:
This breaks https://github.com/sabbelasichon/typo3-rector/pull/4277 again.
can we revert this change or find a Solution for this problem?

simonschaufi:
As you can see, this is actually a fix for another issue. I wonder why you define the path in the cli and not in rector config?

Kanti:
For normal rector execution, I don't need to specify the path in the rector config.
I did not think that I need to define in what path the root of the project is.

I searched the rector.php examples for both rector and typo3-rector and could not find a way to define the path.
How would I define the path?

simonschaufi:
You don't specify the root path but only the paths in which rector should run, see:
https://github.com/sabbelasichon/typo3-rector/blob/main/templates/rector.php.dist#L17
Rector will the figure out the path by itself.

Kanti:
This is a really good piece of info :)

We have a helper in the paths and give all files individually to rector:
````php
$rectorConfig->paths(
    array_filter(explode("\n", (string)shell_exec("git ls-files | xargs ls -d 2>/dev/null | grep -E '\.(php)$'")))
);
````

It works as you said if I just set the path a directory.

````php
$rectorConfig->paths(
    [__DIR__ . '/src/']
);
````

But this is not nice if we only have the option to define paths if we want to use typo3-rector.

This also applies to the rector command if you add a filename as a parameter.
e.g:
`rector process src/extension/test_extension/ext_tables.php`.

Then I still get the error `Unable to create directory at /src/extensions/test_extension/Configuration/TCA/Overrides. mkdir(): Permission denied'. This is also a problem if I only set the directory in the rector config.

@simonschaufi can we have the `getcwd()` back in the code? or should we find another solution?

simonschaufi:
If getcwd breaks the code, then we should find another solution or a real fix that doesn't break but I'm not motivated to spend time in this right now as defining the paths works great.

Kanti:
I have written a summary in the issue #4301
I will try to find a solution without breaking the usage of the CLI from a different Project path.

Issue #4298, opened 2024-08-09
[BUG]: SimplifyCheckboxItemsTCARector removes non-empty labels (TYPO3 11)


LeoniePhiline:
### Minimal PHP Code Causing Issue

```php
<?php

return [
  'columns' => [
     'col_name' => [
         'config'   => [
             'type'       => 'check',
             'renderType' => 'checkboxToggle',
             'items'      => [
                 ['LLL:EXT:some_extension/Resources/Private/Language/locallang_db.xlf:some_label', ''],
             ],
         ],
     ],
  ],
];
```

### Applied rules

`SimplifyCheckboxItemsTCARector`

### Expected Behaviour

```php
<?php

return [
  'columns' => [
     'col_name' => [
         'config'   => [
             'type'       => 'check',
             'renderType' => 'checkboxToggle',
             'items'      => [
                 [0 => 'LLL:EXT:some_extension/Resources/Private/Language/locallang_db.xlf:some_label'],
             ],
         ],
     ],
  ],
];
```

Note: Manually changing `['LLL:EXT:some_extension/Resources/Private/Language/locallang_db.xlf:some_label', '']` to `[0 => 'LLL:EXT:some_extension/Resources/Private/Language/locallang_db.xlf:some_label']` prevents `SimplifyCheckboxItemsTCARector` from triggering and falsely removing the label.

### Package Version

2.6.4

### PHP Version

8.3.10

### TYPO3 Version

11.5.38

### Notes

Actual behavior:

```php
<?php

return [
  'columns' => [
     'col_name' => [
         'config'   => [
             'type'       => 'check',
             'renderType' => 'checkboxToggle',
         ],
     ],
  ],
];
```

The `['items'][...][0]` label gets lost even if it is non-empty. Docs: https://docs.typo3.org/m/typo3/reference-tca/11.5/en-us/ColumnsConfig/Type/Check/Toggle.html#tca_example_checkbox_17

Later, for TYPO3 12, the `0 => ` will be translated into `'label' => `. Docs: https://docs.typo3.org/m/typo3/reference-tca/12.4/en-us/ColumnsConfig/Type/Check/Toggle.html#columns-check-checkboxToggle-examples-17

helsner:
@simonschaufi // @sabbelasichon 
Here it is kinda the question what we want to do.
Basically the original config is wrong, that's why the items config gets killed.
Should we still check for this and prevent a removal?

LeoniePhiline:
[...]

PR #4299, opened 2024-08-29
[BUGFIX] Add numbered key if missing for SimplifyCheckboxItemsTCARector

simonschaufi:
Resolves: #4298


LeoniePhiline:
Thank you very much! 🚀

PR #4315, opened 2024-09-26
[FEATURE] Add MigrateViewHelperRenderStaticRector

s2b:
This rector rule covers the majority of use cases where the now deprecated `renderStatic()` is used in Fluid ViewHelper classes. It's the same rule that has been used in the TYPO3 Core patch that migrated all existing ViewHelpers to `render()`. There were some edge cases in the TYPO3 Core which weren't covered, but those are rather old and don't follow recent best practices.

Core Patch: https://review.typo3.org/c/Packages/TYPO3.CMS/+/85834


sabbelasichon:
@s2b Thank you so much. I have already seen your standalone repository. Great work. I will have a look at it soon.

sabbelasichon:
@simonschaufi I haved fixed all the errors. Could you have a quick look at the rule? I would really like to merge it. Very useful rule

simonschaufi:
I found some more issues and will fix them

simonschaufi (approved):

simonschaufi:
Thank you very much!

PR #4542, opened 2025-05-12
Remove use of protected property NodeNameResolver on Rector Rules

samsonasik:
Rector is removing use of protected property `nodeNameResolver` in rules, use `isName()`, `isNames()`, `getName()` directly instead. Ref PR:

- https://github.com/rectorphp/rector-src/pull/6875


samsonasik:
@simonschaufi Ready to merge 👍

PR #4569, opened 2025-05-30
[BUGFIX] Skip InjectMethodToConstructorInjectionRector for abstract classes

tillhoerner:
In abstract classes method injection should be preferred over constructor injection (see https://docs.typo3.org/m/typo3/reference-coreapi/main/en-us/ApiOverview/DependencyInjection/Index.html#method-injection)


simonschaufi:
Thank you very much for your contribution. Could you please also create a new test case for this? It is very important that every bug is covered with a test.

PR #4603, opened 2025-06-18
[BUGFIX] Move composer-patches to require-dev and update docs

### LINKED ISSUE #4602: [BUG] Requirement of cweagans/composer-patches breaks installations
### Minimal PHP Code Causing Issue

The change https://github.com/sabbelasichon/typo3-rector/commit/b6ca48ced00ad3c2612c544913e65a23bc51f1ac and the following release breaks all installations which use this package due to the new requirement `cweagans/composer-patches`:

- https://github.com/pagemachine/typo3-formlog/actions/runs/15674347933/job/44151183172
- https://github.com/pagemachine/typo3-phinx/actions/runs/15674451130/job/44151491457
- https://github.com/pagemachine/typo3-matomo-tracking/actions/runs/15674234112/job/44151150651

The error is always the same:

```
cweagans/composer-patches contains a Composer plugin which is blocked by your allow-plugins config. You may add it to the list if you consider it safe.                                                                            

You can run "composer config --no-plugins allow-plugins.cweagans/composer-patches [true|false]"
to enable it (true) or disable it explicitly and suppr ess this exception (false)                                                   

See https://getcomposer.org/allow-plugins
```

### Applied rules

none

### Expected Behaviour

Composer installation should work just like before, without allowing additional plugins.

### Package Version

2.14.3

### PHP Version

8.x

### TYPO3 Version

any

### Notes

_No response_

simonschaufi:
Resolves #4602

### DIFF
--- a/docs/all_rectors_overview.md
+++ b/docs/all_rectors_overview.md
@@ -100,7 +100,24 @@ Refactor file ext_emconf.php
 
 ### GeneralUtilityMakeInstanceToConstructorPropertyRector
 
-Move GeneralUtility::makeInstance calls to constructor injection
+Move GeneralUtility::makeInstance calls to constructor injection.
+
+> [!IMPORTANT]
+> If you use this rule, please install the [cweagans/composer-patches](https://packagist.org/packages/cweagans/composer-patches) package and add the following snippet to your composer.json:
+
+```json
+{
+    "extra": {
+        "patches": {
+            "rector/rector": [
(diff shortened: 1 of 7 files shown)


docs/all_rectors_overview.md
```diff
@@ -100,7 +100,30 @@ Refactor file ext_emconf.php
...
+        }
+    }
+}
+```
+
+and execute:
+
+```bash
+composer config --no-plugins allow-plugins.cweagans/composer-patches true
+```
```
mbrodala, line 126: This can be dropped since `cweagans/composer-patches` must be installed anyways, so the plugin will be allowed.

mbrodala (changes requested):

PR #4631, opened 2025-09-02
[TASK] Insert TODO comment to stay consistent with other rules

s2b:
Other rector rules already use `TODO` to annotate manual tasks that need to be done after the rector process. This has the advantage that these will be highlighted and/or listed by supporting IDEs.


s2b:
Maybe even link to docs here? I think this would be the correct page: https://docs.typo3.org/permalink/t3tca:confval-inline-foreign-match-fields

simonschaufi:
Thank you!

PR #4657, opened 2025-09-25
[FEATURE] Add rule to automatically add TCA default values

linawolf:
With TYPO3 13 the ext_tables.sql file is not mandatory anymore.
When removing the ext_tables.sql TCA fields without a default value can
cause errors when thawing Extbase model objects

(Fallback chaing default value of the model property -> TCA default value -> ext_tables.sql default value). Having a default value for all TCA fields that are likely mapped to models can prevent errors on thawing.

* Existing default values are preserved.

* If the field is nullable, the default value is null

* For string fields (input, color, email, text, ...) the default is an empty string

* For numeral fields (number, select, group, ...) the default is zero

### DIFF
--- a/rules/TYPO313/v4/TcaDefaultsRector.php
+++ b/rules/TYPO313/v4/TcaDefaultsRector.php
@@ -1,0 +1,97 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\TYPO313\v4;
+
+use PhpParser\Node\Expr;
+use PhpParser\Node\Expr\Array_;
+use PhpParser\Node\Expr\ArrayItem;
+use PhpParser\Node\Expr\ConstFetch;
+use PhpParser\Node\Name;
+use PhpParser\Node\Scalar\Int_;
+use PhpParser\Node\Scalar\String_;
+use Ssch\TYPO3Rector\Contract\NoChangelogRequiredInterface;
+use Ssch\TYPO3Rector\Rector\AbstractTcaRector;
+use Symplify\RuleDocGenerator\Contract\DocumentedRuleInterface;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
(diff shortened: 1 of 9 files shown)


rules/TYPO313/v4/TcaDefaultsRector.php
```diff
@@ -0,0 +1,99 @@
...
+use PhpParser\Node\Name;
+use PhpParser\Node\Scalar\Int_;
+use PhpParser\Node\Scalar\String_;
+use Ssch\TYPO3Rector\Rector\AbstractTcaRector;
+use Symplify\RuleDocGenerator\Contract\DocumentedRuleInterface;
+use Symplify\RuleDocGenerator\ValueObject\CodeSample\CodeSample;
+use Symplify\RuleDocGenerator\ValueObject\RuleDefinition;
+
+/**
+ * @changelog
```
simonschaufi, line 20: Since there is no changelog, an additional interface needs to be used: NoChangelogRequiredInterface

rules/TYPO313/v4/TcaDefaultsRector.php
```diff
@@ -0,0 +1,99 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\TYPO313\v4;
```
simonschaufi, line 5: Is this really the case since TYPO3 13.4 or should this rule not go into 13.0?

rules/TYPO313/v4/TcaDefaultsRector.php
```diff
@@ -0,0 +1,99 @@
...
+            || $this->isConfigType($configArray, 'radio')
+            || $this->isConfigType($configArray, 'inline')
+            || $this->isConfigType($configArray, 'select')
+            || $this->isConfigType($configArray, 'group')
+        ) {
+            $configArray->items[] = new ArrayItem(new Int_(0), new String_('default'));
+            $this->hasAstBeenChanged = true;
+            return;
+        }
+
```
simonschaufi, line 97: remove empty line

PR #4668, opened 2025-10-03
[FEATURE] Add ReplaceLocationParsersWithLoadersRector

tomasnorre:
I have at least managed to get the tests added, to prove that the Rector is still failing. 

Boah it has been a long time since I wrote my last Rector rule, would be nice with a pointer or an example of another Rector that I can peek into for inspiration to reach the goal, thought I picked an easy one :)

### DIFF
--- a/rules/TYPO314/v0/ReplaceLocalizationParsersWitHLoaders.php
+++ b/rules/TYPO314/v0/ReplaceLocalizationParsersWitHLoaders.php
@@ -1,0 +1,53 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Ssch\TYPO3Rector\TYPO314\v0;
+
+use PhpParser\Node;
+use PhpParser\Node\Expr\MethodCall;
+use Rector\Rector\AbstractRector;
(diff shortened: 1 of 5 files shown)


rules/TYPO314/v0/ReplaceLocalizationParsersWitHLoaders.php
```diff
@@ -0,0 +1,53 @@
...
+            ]
+        );
+    }
+
+    /**
+     * @return array<class-string<Node>>
+     */
+    public function getNodeTypes(): array
+    {
+        return [MethodCall::class];
```
simonschaufi, line 43: This should probably be Assignment

rules/TYPO314/v0/ReplaceLocalizationParsersWitHLoaders.php
```diff
@@ -0,0 +1,53 @@
...
+    }
+
+    /**
+     * @param MethodCall $node
+     */
+    public function refactor(Node $node): Node
+    {
+        return $node;
```
simonschaufi, line 51: #4667 might help you. I used Gemini to do the main logic and then finished the rule myself.

Issue #4708, opened 2025-10-29
Deprecation: #107537 - GeneralUtility::createVersionNumberedFilename


simonschaufi:
Deprecation: #107537 - GeneralUtility::createVersionNumberedFilename

https://docs.typo3.org/c/typo3/cms-core/main/en-us/Changelog/14.0/Deprecation-107537-CreateVersionNumberedFileName.html

# Deprecation: \#107537 - GeneralUtility::createVersionNumberedFilename

See [107537](https://forge.typo3.org/issues/107537)

## Description

<span class="title-ref">GeneralUtility::createVersionNumberedFilename</span>
adds cache busting to a file URL, when called in a certain and correct
order with other legacy API methods to create URLs from system
resources.

This class and its functionality is superseded by the **System Resource
API**.

## Impact

Calling this method will trigger a PHP deprecation warning. The method
will continue to work as is, until it is removed in TYPO3 v15.0.

## Affected installations

TYPO3 installations with custom extensions/code that directly call this
deprecated method:

- `GeneralUtility::createVersionNumberedFilename`

## Migration

Use the **System Resource API** instead.

Before:

``` php
public function renderUrl(string $file): string
{
    $file = GeneralUtility::getFileAbsFileName($file);
    $partialUrl = GeneralUtility::createVersionNumberedFilename($file);
    return PathUtility::getAbsoluteWebPath($partialUrl);
}
```

After:

``` php
public function __construct(
    private readonly SystemResourceFactory $systemResourceFactory,
    private readonly SystemResourcePublisherInterface $resourcePublisher,
) {}

public function renderUrl(string $resourceIdentifier, ServerRequestInterface $request): string
{
    $resource = $this->systemResourceFactory->createPublicResource($resourceIdentifier);
    return (string)$this->resourcePublisher->generateUri(
        $resource,
        $request,
        new UriGenerationOptions(absoluteUri: true),
    );
}
```

<div class="index">

PHP-API, FullyScanned, ext:core

</div>

PR #4756, opened 2025-11-21
[BUGFIX] Handle return case in MigrateGeneralUtilityCreateVersionNumberedFilenameRector

simonschaufi:
Resolves: #4708


simonschaufi:
cc @tomasnorre see my fix here

PR #4878, opened 2026-04-13
Include rector files

chrisprog20:
It seems that in this file the inclusion of the files v14/tca-140.php and v14/typo3-14.2.php is lacking. This commit adds them.


simonschaufi:
thank you!

PR #4884, opened 2026-05-11
[BUGFIX] Guard against null dim in MoveSchedulerFrequencyOptionsToTCARector

### LINKED ISSUE #4882: MoveSchedulerFrequencyOptionsToTCARector crashes with null dim on non-subscript array assignments
## Description

`MoveSchedulerFrequencyOptionsToTCARector` throws a fatal error when processing PHP files that contain array assignments using the append operator (`[]`) rather than an explicit subscript key. The `->dim` property of `ArrayDimFetch` is `null` in that case, which is valid PHP-Parser AST, but `shouldSkip()` passes it directly to `ValueResolver::isValue()` without a null guard.

## Error

```
System error: "Rector\PhpParser\Node\Value\ValueResolver::isValue():
Argument #1 ($expr) must be of type PhpParser\Node\Expr, null given,
called in .../ssch/typo3-rector/rules/TYPO314/v0/MoveSchedulerFrequencyOptionsToTCARector.php:147"
```

## Root Cause

In `MoveSchedulerFrequencyOptionsToTCARector::shouldSkip()` (line ~147):

```php
$freqOptionsDimFetch = $assignVar->var;
if (! $freqOptionsDimFetch instanceof ArrayDimFetch
    || ! $this->valueResolver->isValue($freqOptionsDimFetch->dim, 'frequencyOptions')
) {
    return true;
}
```

`ArrayDimFetch::$dim` is typed as `?Expr` — it is `null` when the array uses the append syntax (`$arr[] = ...`). The guard checks `instanceof ArrayDimFetch` but does not verify `$freqOptionsDimFetch->dim !== null` before passing it to `isValue()`, which requires a non-null `Expr`.

## Fix

Add a null check before calling `isValue()`:

```php
if (! $freqOptionsDimFetch instanceof ArrayDimFetch
    || $freqOptionsDimFetch->dim === null
    || ! $this->valueResolver->isValue($freqOptionsDimFetch->dim, 'frequencyOptions')
) {
    return true;
}
```

The same pattern applies to the subsequent `$schedulerDimFetch->dim` and `$scOptionsDimFetch->dim` checks in the same method.

## Reproduction

Any PHP file processed under the `TYPO3_14` / `UP_TO_TYPO3_14` rule set that contains a method or function with an append-syntax assignment on line 85 (or nearby lines covered by the rule's node visitor) will trigger the crash. The files in our project that trigger it are unrelated to Scheduler — they are generic PHP classes that happen to contain `$data[] = ...` patterns.

## Environment

- ssch/typo3-rector: v3.14.1
- Rector: 2.4.2
- PHP: 8.4.19
- Rule set used: `Typo3LevelSetList::UP_TO_TYPO3_14`

dwenzel:
## Summary

Fixes #4882.

`MoveSchedulerFrequencyOptionsToTCARector::shouldSkip()` passes `ArrayDimFetch::$dim` directly to `ValueResolver::isValue()` without checking for `null` first. `$dim` is typed `?Expr` and is `null` whenever the append operator (`[]`) is used on the left-hand side of an assignment. A nested append pattern such as `$data[][] = 'value'` produces an inner `ArrayDimFetch` with `dim === null`, which causes a fatal type error.

- Add `$xxx->dim === null` guards before each `isValue($xxx->dim, ...)` call in `shouldSkip()` (four sites: `frequencyOptions`, `scheduler`, `SC_OPTIONS`, `TYPO3_CONF_VARS`)
- Add a test fixture (`Fixture/append_syntax/ext_localconf.php.inc`) with the triggering pattern and a dedicated test method that asserts the file passes through unchanged without throwing

### DIFF
--- a/tests/Rector/v14/v0/MoveSchedulerFrequencyOptionsToTCARector/Fixture/extension2/ext_localconf.php.inc
+++ b/tests/Rector/v14/v0/MoveSchedulerFrequencyOptionsToTCARector/Fixture/extension2/ext_localconf.php.inc
@@ -3,21 +3,10 @@
 namespace Ssch\TYPO3Rector\Tests\Rector\v13\v0\MoveSchedulerFrequencyOptionsToTCARector\Fixture;
 
 $GLOBALS['TYPO3_CONF_VARS']['SC_OPTIONS']['scheduler']['frequencyOptions']['0 1 * * *'] = 'LLL:EXT:my_extension/Resources/Private/Language/locallang.xlf:daily_2am';
-
-$data[]['key'] = 'value';
-$data[]['frequencyOptions']['key'] = 'value';
-$data[]['scheduler']['frequencyOptions']['key'] = 'value';
-$data[]['SC_OPTIONS']['scheduler']['frequencyOptions']['key'] = 'value';
-
 ?>
 -----
 <?php
 
 namespace Ssch\TYPO3Rector\Tests\Rector\v13\v0\MoveSchedulerFrequencyOptionsToTCARector\Fixture;
-
-$data[]['key'] = 'value';
-$data[]['frequencyOptions']['key'] = 'value';
-$data[]['scheduler']['frequencyOptions']['key'] = 'value';
-$data[]['SC_OPTIONS']['scheduler']['frequencyOptions']['key'] = 'value';
 
 ?>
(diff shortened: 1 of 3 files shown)


simonschaufi:
Fixed with #4915

PR #4887, opened 2026-05-21
Add StructArmed to QA

samsonasik:
This PR adds StructArmed to QA

https://github.com/boundwize/structarmed

For starter, use PSR4 preset, and already catch some PSR4 violations and already included in this PR.

`StructArmed` requires 8.2 so I only add on github workflow.


samsonasik:
All green 🎉  @simonschaufi ready 👍

simonschaufi:
Thank you very much! I actually wanted to do the same but you were faster ;)

PR #4890, opened 2026-06-01
[BUGFIX] Skip HashService calls with an unexpected argument count

astehlik:
The argument count check only rejected calls with more arguments than expected, not fewer. A validateHmac() call with a single argument (instead of the required two) slipped through and hit $node->args[1], which does not exist, causing a crash. 

Replace the two ">" bound checks with a single exact-count check per method (2 for validateHmac(), 1 for the others) and add fixtures covering an unrelated class that happens to declare its own validateHmac() method, and a validateHmac() call with too few arguments.

### DIFF
--- a/tests/Rector/v13/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector/Fixture/skip_validate_hmac_with_unexpected_argument_count.php.inc
+++ b/tests/Rector/v13/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector/Fixture/skip_validate_hmac_with_unexpected_argument_count.php.inc
@@ -1,0 +1,25 @@
+<?php
+
+namespace Ssch\TYPO3Rector\Tests\Rector\v13\v0\MigrateExtbaseHashServiceToUseCoreHashServiceRector\Fixture;
+
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Security\Cryptography\HashService;
+
+$hashService = GeneralUtility::makeInstance(HashService::class);
+
+$hashService->validateHmac('123');
+
+?>
+-----
+<?php
+
+namespace Ssch\TYPO3Rector\Tests\Rector\v13\v0\MigrateExtbaseHashServiceToUseCoreHashServiceRector\Fixture;
+
+use TYPO3\CMS\Core\Crypto\HashService;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+$hashService = GeneralUtility::makeInstance(HashService::class);
+
+$hashService->validateHmac('123');
+
+?>
--- a/rules/TYPO313/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector.php
+++ b/rules/TYPO313/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector.php
@@ -84,17 +84,18 @@ CODE_SAMPLE
             return null;
         }
 
-        if (count($node->args) > 1 && ! $this->isName($node->name, 'validateHmac')) {
-            return null;
-        }
+        // The Extbase signatures are validateHmac($input, $hmac) (2 arguments) and a single $input
+        // argument for the other methods. Skip if the number of arguments does not match.
+        $isValidateHmac = $this->isName($node->name, 'validateHmac');
 
-        if (count($node->args) > 2 && $this->isName($node->name, 'validateHmac')) {
+        $expectedArgumentCount = $isValidateHmac ? 2 : 1;
+        if (count($node->args) !== $expectedArgumentCount) {
             return null;
         }
 
         $additionalSecretArgument = $this->nodeFactory->createArg($this->additionalSecret);
 
-        if ($this->isName($node->name, 'validateHmac')) {
+        if ($isValidateHmac) {
             $node->args[2] = $node->args[1];
             $node->args[1] = $additionalSecretArgument;
         } else {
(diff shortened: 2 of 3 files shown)


tests/Rector/v13/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector/Fixture/skip_validate_hmac_with_unexpected_argument_count.php.inc
```diff
@@ -0,0 +1,25 @@
...
+<?php
+
+namespace Ssch\TYPO3Rector\Tests\Rector\v13\v0\MigrateExtbaseHashServiceToUseCoreHashServiceRector\Fixture;
+
+use TYPO3\CMS\Core\Crypto\HashService;
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+
+$hashService = GeneralUtility::makeInstance(HashService::class);
+
+$hashService->validateHmac('123');
```
simonschaufi, line 23: something with this line is not right because `validateHmac` from the core needs an additional secret and hmac argument. The test case is wrong in my opinion.

simonschaufi:
The wrong class name is fixed in #4892, I'm not sure about the rest

tests/Rector/v13/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector/Fixture/skip_validate_hmac_with_unexpected_argument_count.php.inc
astehlik: I refactored the number of argument checks a bit to better understand what is happening.

To make sure the change doesn't beak anything, the test is added. It checks, that refactoring is skipped, when the number of arguments does not match.

But I would also be fine with your simpler approach :slightly_smiling_face:

simonschaufi: I would gladly merge your change but the migrated code is broken.

simonschaufi:
@astehlik can you please have a look at your PR again? Otherwise I'll close it

astehlik:
@simonschaufi If you can tell me, which changes are needed, and point me to the error that occurs, I will take another look.

astehlik:
@simonschaufi I've realized there is another issue that is fixed with this PR and now covered by the test.

Please let me know if it works for you :slightly_smiling_face:

tests/Rector/v13/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector/Fixture/skip_validate_hmac_with_unexpected_argument_count.php.inc
simonschaufi: I need a test case where the extbase version is migrated to the core version but as I said, the core needs an additional secret. That's why your test case is broken currently.

astehlik: @simonschaufi Maybe there is a misunderstanding. I've added a test case to check the skip logic in the [MigrateExtbaseHashServiceToUseCoreHashServiceRector](https://github.com/sabbelasichon/typo3-rector/blob/main/rules/TYPO313/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector.php#L87-L93)

Therefore my test code has intentionally the wrong number of arguments. The test detected, that the skip logic does not always work.

I hope this clarifies it :slightly_smiling_face:

simonschaufi: I wonder why the code should be skipped but a migration is still done (from TYPO3\CMS\Extbase\Security\Cryptography\HashService to TYPO3\CMS\Core\Crypto\HashService). A real skip should not migrate anything.

tests/Rector/v13/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector/Fixture/skip_validate_hmac_with_unexpected_argument_count.php.inc
```diff
@@ -0,0 +1,25 @@
+<?php
+
+namespace Ssch\TYPO3Rector\Tests\Rector\v13\v0\MigrateExtbaseHashServiceToUseCoreHashServiceRector\Fixture;
+
+use TYPO3\CMS\Core\Utility\GeneralUtility;
+use TYPO3\CMS\Extbase\Security\Cryptography\HashService;
+
+$hashService = GeneralUtility::makeInstance(HashService::class);
+
+$hashService->validateHmac('123');
```
simonschaufi, line 10: I also wonder about this example because extbase HashService->validateHmac requires 2 parameters:

```php
public function validateHmac(string $string, string $hmac): bool
```

so the original code is already invalid. Can you please explain why your given input code is intentionally invalid?

astehlik: @simonschaufi My idea was to add a test that checks if the conditions in [MigrateExtbaseHashServiceToUseCoreHashServiceRector](https://github.com/sabbelasichon/typo3-rector/blob/main/rules/TYPO313/v0/MigrateExtbaseHashServiceToUseCoreHashServiceRector.php#L87-L93) work as expected.

The checks made sense since in older versions when it was not fully safe, that the HashService is the correct one. Since this was fixed, I'm not sure how relevant this scenario is and if we should not throw an Exception instead, when the number for arguments does not match. Because the client's code would be invalid and it might make more sense to notify them about it instead of silently skipping.

simonschaufi: The method always had 2 arguments since the beginning: https://github.com/TYPO3/typo3/commit/876f348f6cc79ff66bca2b91f83559e6bc2434e8 Therefore your code is invalid and I will close this PR unmerged. Rector's responsibility is not to fix broken code but to get valid code as input and migrate that. Sorry. To fix your code first, use tools like phpstan.

PR #4921, opened 2026-08-21
StructArmed: Enable ExtendedClassMustBeAbstractOrInstantiatedRule

samsonasik:
StructArmed 0.16 bring YAGNI preset, and one of its rule is: `ExtendedClassMustBeAbstractOrInstantiatedRule`.

The rule is check:

- class is only extended
- class never instantiated

Then class needs to be marked as `abstract`. For safety purpose, this PR register this rule with only apply on `classNamePattern` with `Abstract` prefix in it:

```php
    ->rule(
        'yagni.extended_class.must_be_abstract_or_instantiated',
        new ExtendedClassMustBeAbstractOrInstantiatedRule(
            layer: 'Source',
            classNamePattern: '/^TYPO3\\\\.*\\\\Abstract.*$/',
        ),
    )
```

and applied change is only on `stubs` so should be safer to apply. 

This applied automatically with `--fix` command:

```
➜  typo3-rector git:(main) ✗ vendor/bin/structarmed analyze --fix
Analyzing [============================] 100% 1370/1370

✓  5 violations have been fixed.
```

### DIFF
--- a/structarmed.php
+++ b/structarmed.php
@@ -4,12 +4,20 @@ declare(strict_types=1);
 
 use Boundwize\StructArmed\Architecture;
 use Boundwize\StructArmed\Preset\Preset;
+use Boundwize\StructArmed\Rule\Rules\Class_\ExtendedClassMustBeAbstractOrInstantiatedRule;
 use Boundwize\StructArmed\Rule\Rules\Class_\MustBeFinalRule;
 
 return Architecture::define()
     ->rule(
         'source.must_be_final',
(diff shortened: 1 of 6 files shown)


samsonasik:
@simonschaufi ready to merge 👍

simonschaufi:
This is not adding any useful rule to the code. The test stubs are identical to the TYPO3 core and there is no need to declare these classes abstract if the TYPO3 core classes are not abstract. It would be a different story if the Rector rules would be changed which I would agree on but the stubs are irrelevant.

PR #4931, opened 2026-09-13
[FEATURE] Add MigrateExtensionManagementUtilityAddTcaSelectItemToAddPluginRector

simonschaufi:
I created this rule to migrate actual plugin registrations like here: https://github.com/benjaminkott/bootstrap_package/blob/76f079a85e998748d42c9d01772e92e4efd53b06/Configuration/TCA/Overrides/200_content_element_accordion.php#L18-L28 and `addPlugin` is the official API for registering non extbase content elements, see https://docs.typo3.org/permalink/t3coreapi:plugins-non-extbase and it has the advantage that FlexForms can also be added since TYPO3 14.

PS: The core should really be cleaned up, especially this class as there are so many methods in there that do very similar things and there is a lot of duplicated code in there.


chrissonntag:
Shouldn't it rather be `addRecordType()` instead of `addPlugin()`?
See https://api.typo3.org/main/classes/TYPO3-CMS-Core-Utility-ExtensionManagementUtility.html#method_addRecordType

simonschaufi:
See description :smile:

